PR #26, opened 2023-12-09
Add parser

tomekzaw:


parser/index.js
```diff
@@ -0,0 +1,164 @@
...
+  function appendSyntax(syntax) {
+    addChildrenWithStyle(syntax, 'syntax');
+  }
+
+  function addChildrenWithStyle(node, style) {
+    const start = text.length;
+    processChildren(node);
+    const end = text.length;
+    ranges.push([style, start, end - start]);
+  }
```
s77rt, line 83: Hi, I think this caused a regression, Although we are processing a given node (e.g. `italic`), its range is added after `processChildren` in which we will add other ranges too. This process forces us to sort the ranges. However the current sorting algorithm does not take into account ranges that have same start and length attributes, so we won't swap them in cases where we should because they were originally processed in reverse.

I think a better approach was to add the range then `processChildren` and then change the range length, but it may have been too late for that approach as we'd have to fix some failing tests. See more here https://github.com/Expensify/App/issues/62029#issuecomment-2906972031

Anyway, we opted for a simple solution which is to add an edge case to handle when start and length are the same and swapping is needed (which is so far only encountered with emojis)

PR #32, opened 2023-12-13
Add web implementation

Skalakid:


tomekzaw:

Skalakid:

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+  defaultInputStyles: {
+    borderColor: 'black',
+    borderWidth: 1,
+    borderStyle: 'solid',
+    fontFamily: 'sans-serif',
+    boxSizing: 'border-box',
+    whiteSpace: 'pre-wrap',
+    overflow: 'scroll',
+    scrollbarWidth: 'none',
+    paddingHorizontal: 100,
```
tomekzaw, line 434: Why do we need it?

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+      value,
+    },
+    ref,
+  ) => {
+    const divRef = useRef<HTMLDivElement | null>(null);
+    const currentlyFocusedField = useRef<HTMLDivElement | null>(null);
+    const contentSelection = useRef<Selection | null>(null);
+    const className = `react-native-live-markdown-input-${multiline ? 'multiline' : 'single-line'}`;
+    const history = useRef(new MarkdownInputHistory(20));
+    const flattenStyle = useMemo(() => StyleSheet.flatten(style), [style]);
```
tomekzaw, line 110:
```suggestion
    const flattenedStyle = useMemo(() => StyleSheet.flatten(style), [style]);
```

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+      markdownStyle,
+      onBlur,
+      onChange,
+      onChangeText,
+      onClick,
+      onFocus,
+      onKeyPress,
+      onSelectionChange,
+      onSubmitEditing,
+      placeholder = '\u200B', // Zero-width space to prevent div from collapsing
```
tomekzaw, line 96: What if someone passes `placeholder=""`? Maybe we should insert zero-width space when rendering the placeholder instead of here?

Also, let's move this string to a const:
```ts
const ZERO_WIDTH_SPACE = '\u200B';
```

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+  start: number;
+  end: number;
+};
+
+let focusTimeout: NodeJS.Timeout | null = null;
+
+// Removes one '\n' from the end of the string that were added by contentEditable div
+const normalizeValue = (value: string) => {
+  return value.replace(/\n$/, '');
+};
```
tomekzaw, line 65: Can we convert all arrow functions into function declarations? 
```suggestion
function normalizeValue(value: string) {
  return value.replace(/\n$/, '');
}
```

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+
+        let text = '';
+        const nativeEvent = e.nativeEvent as MarkdownNativeEvent;
+        if (nativeEvent.inputType === 'historyUndo') {
+          text = undo(divRef.current);
+        } else if (nativeEvent.inputType === 'historyRedo') {
+          text = redo(divRef.current);
+        } else {
+          text = parseText(divRef.current, e.target.innerText, processedMarkdownStyle).text;
+        }
```
tomekzaw, line 189: Can we use `switch/case` instead?

src/web/parserUtils.ts
```diff
@@ -0,0 +1,208 @@
...
+      Object.assign(node.style, {
+        ...markdownStyle.link,
+        textDecoration: 'underline',
+      });
+      break;
+    case 'code':
+      Object.assign(node.style, {
+        ...markdownStyle.code,
+        fontFamily: 'monospace',
+      });
```
tomekzaw, line 45: `markdownStyle.code` already includes `fontFamily` with the default value of `monospace` on Web.
```suggestion
      Object.assign(node.style, markdownStyle.code);
```

src/web/parserUtils.ts
```diff
@@ -0,0 +1,208 @@
...
+      break;
+    case 'pre':
+      Object.assign(node.style, {
+        ...markdownStyle.pre,
+        fontFamily: 'monospace',
+      });
```
tomekzaw, line 51: `markdownStyle.code` already includes `fontFamily` with the default value of 
```suggestion
      Object.assign(node.style, markdownStyle.pre);
```

src/web/parserUtils.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  const stack = [...ranges];
+  const nestedStack: NestedNode[] = [{node: root, endIndex: textLength}];
+  let lastRangeEndIndex = 0;
+  while (stack.length > 0) {
+    const range = stack.shift();
+    if (!range) {
+      break;
+    }
+    // const [name, index, rangeLength] = range;
```
tomekzaw, line 112:
```suggestion
```

src/web/cursorUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+function getCurrentCursorPosition(target: HTMLElement) {
+  const selection = document.getSelection();
+  if (!selection || selection.rangeCount === 0) {
+    return {start: target.innerText.length, end: target.innerText.length};
+  }
+
+  const range = selection.getRangeAt(0);
+  const startPosition = getIndexedPosition(target, range, true);
+  const endPosition = getIndexedPosition(target, range, false);
+  return {start: startPosition, end: endPosition};
```
tomekzaw, line 88:
```suggestion
  const start = getIndexedPosition(target, range, true);
  const end = getIndexedPosition(target, range, false);
  return {start, end};
```

src/web/cursorUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+
+function getIndexedPosition(target: HTMLElement, range: Range, isStart: boolean) {
+  const marker = document.createTextNode('\0');
+  const rangeClone = range.cloneRange();
+
+  if (isStart) {
+    rangeClone.collapse(true);
+  } else {
+    rangeClone.collapse(false);
+  }
```
tomekzaw, line 68:
```suggestion
  rangeClone.collapse(isStart);
```

src/web/MarkdownTextInput.css
```diff
@@ -0,0 +1,25 @@
+.react-native-live-markdown-input-single-line[contenteditable='true'] {
```
tomekzaw, line 1: Looks like [here](https://www.w3schools.com/css/css_attribute_selectors.asp) they use `"` instead of `'` so let's change it everywhere in this file:
```suggestion
.react-native-live-markdown-input-single-line[contenteditable="true"] {
```

src/MarkdownTextInput.web.tsx
tomekzaw, line 1: Let's introduce the following renames in a separate PR, once this one is merged:
* `MarkdownTextInput.tsx` &rarr; `MarkdownTextInput.native.tsx`
* `MarkdownTextInput.web.tsx` &rarr; `MarkdownTextInput.tsx`

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+import * as ParseUtils from './web/parserUtils';
+import * as CursorUtils from './web/cursorUtils';
+import * as StyleUtils from './styleUtils';
+import type * as MarkdownTextInputDecoratorViewNativeComponent from './MarkdownTextInputDecoratorViewNativeComponent';
+import './web/MarkdownTextInput.css';
+import MarkdownInputHistory from './web/historyClass';
 
-import React from 'react';
-import type {TextInputProps} from 'react-native';
+// TODO: remove require
```
tomekzaw, line 22:
```suggestion
```

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+}
+
+type MarkdownStyle = MarkdownTextInputDecoratorViewNativeComponent.MarkdownStyle;
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
 interface MarkdownTextInputProps extends TextInputProps {
-  // nothing here
+  markdownStyle?: MarkdownStyle;
+  onClick?: (e: MouseEvent<HTMLDivElement>) => void;
+  dir?: string;
```
tomekzaw, line 48: Why do we need to explicitely add these props here? Aren't they already defined in `TextInputProps`?

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+}
+
+interface MarkdownNativeEvent extends Event {
```
tomekzaw, line 51: We should find a better name for this event type

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+      onBlur,
+      onChange,
+      onChangeText,
+      onClick,
+      onFocus,
+      onKeyPress,
+      onSelectionChange,
+      onSubmitEditing,
+      placeholder = '\u200B', // Zero-width space to prevent div from collapsing
+      placeholderTextColor = '#ccc',
```
tomekzaw, line 97: Let's use a semi-transparent black color here, i.e. `rgba(0,0,0,0.2)`, so it works with different background colors as well.

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+      spellCheck,
+      style = {},
+      value,
+    },
+    ref,
+  ) => {
+    const divRef = useRef<HTMLDivElement | null>(null);
+    const currentlyFocusedField = useRef<HTMLDivElement | null>(null);
+    const contentSelection = useRef<Selection | null>(null);
+    const className = `react-native-live-markdown-input-${multiline ? 'multiline' : 'single-line'}`;
```
tomekzaw, line 108: Let's unify the class names here
```suggestion
    const className = `react-native-live-markdown-input-${multiline ? 'multiline' : 'singleline'}`;
```

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+    const history = useRef(new MarkdownInputHistory(20));
```
tomekzaw, line 109: This syntax will create a new `MarkdownInputHistory` object on each call, how about this?
```suggestion
    const history = useRef(null);
    if (history.current === null) {
      history.current = new MarkdownInputHistory(20);
    }
```

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+        const parsedText = ParseUtils.parseText(target, text, cursorPosition, customMarkdownStyles, !multiline);
+        if (history.current && shouldAddToHistory) {
+          history.current.debouncedAdd(parsedText.text, parsedText.cursorPosition);
+        }
+        return parsedText;
+      },
+      [multiline],
+    );
+
+    const processedMarkdownStyle = React.useMemo(() => {
```
tomekzaw, line 125: Let's use `useMemo` directly instead of `React.useMemo` (don't forget to apply this suggestion in other places as well).

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+
+    const redo = useCallback(
+      (target: HTMLDivElement) => {
+        const item = history.current.redo();
+        return parseText(target, item ? item.text : null, processedMarkdownStyle, item ? item.cursorPosition : null, false).text;
+      },
+      [processedMarkdownStyle],
+    );
+
+    // We have to process value porperty since contentEditable div adds one additional '\n' at the end of the text if we are entering new line
```
tomekzaw, line 149:
```suggestion
    // We have to process value property since contentEditable div adds one additional '\n' at the end of the text if we are entering new line
```

src/MarkdownTextInput.web.tsx
```diff
@@ -1,22 +1,438 @@
...
+          (e.target as HTMLInputElement).value = text;
+        }
+        e.nativeEvent.text = text;
+      }
+      return e;
+    }, []);
+
+    // Placeholder text color logic
+    const updateTextColor = useCallback((target: HTMLDivElement, text: string) => {
+      const node = target;
```
tomekzaw, line 171:
```suggestion
    const updateTextColor = useCallback((node: HTMLDivElement, text: string) => {
```

src/styleUtils.ts
```diff
@@ -0,0 +1,102 @@
+import {processColor, Platform} from 'react-native';
+import type * as MarkdownTextInputDecoractorView from './MarkdownTextInputDecoratorViewNativeComponent';
+
+type MarkdownStyle = MarkdownTextInputDecoractorView.MarkdownStyle;
+
+type PartialMarkdownStyle = Partial<{
+  [K in keyof MarkdownStyle]: Partial<MarkdownStyle[K]>;
+}>;
```
tomekzaw, line 8: Let's add a newline here
```suggestion
}>;

```

src/styleUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (input !== undefined) {
+    Object.keys(input).forEach((key) => {
+      if (!(key in output)) {
+        return;
+      }
+      Object.assign(output[key as keyof MarkdownStyle], input[key as keyof MarkdownStyle]);
+    });
+  }
+
+  if (shouldAddUnitsToNumberValues) {
```
tomekzaw, line 64: I don't like that this function does two things at once, can we move adding units to a separate function `processUnitsInMarkdownStyle`? Also, there should be some function from `react-native-web` that already does that (see how `processColor` is used below)

tomekzaw (changes requested):

src/MarkdownTextInput.web.tsx
robertKozik: leftovers after testing 😅 removed 👌🏼

src/MarkdownTextInput.web.tsx
robertKozik: It looks like they're not present inside `TextInputProps`.

types/types.ts
```diff
@@ -0,0 +1,9 @@
+type MarkdownType = 'bold' | 'italic' | 'strikethrough' | 'link' | 'code' | 'pre' | 'blockquote' | 'h1' | 'syntax' | 'mention-here' | 'mention-user';
+
+type MarkdownRange = {
+  type: MarkdownType;
+  startIndex: number;
+  length: number;
+};
+
+export type {MarkdownType, MarkdownRange};
```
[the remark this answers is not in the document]
robertKozik: Actually I don't see any usage inside `parser` directory. I'll move it inside `src/web/parserUtlis.ts` as it's only used there and in test directory

src/MarkdownTextInputDecoratorViewNativeComponent.ts
```diff
@@ -3,6 +3,8 @@ import type {ColorValue, ViewProps} from 'react-native';
 import type {Float} from 'react-native/Libraries/Types/CodegenTypes';
 import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 
+type NumberValue = Float | string;
```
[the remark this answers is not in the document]
Skalakid: You are right, we can proceed without this change. We don't want to allow user passing styles with web units. I removed this change

src/web/MarkdownTextInput.css
Skalakid: Prettier changes it to single quote `'` after saving

tomekzaw (approved):
@Skalakid @robertKozik Congratulations on this PR! 🥳

tomekzaw (approved):

tomekzaw (approved):

src/MarkdownTextInput.web.tsx
tomekzaw: See #168

PR #120, opened 2024-01-11
Remove patching class ExpensiMark with an object

### LINKED ISSUE #41: Remove patching class ExpensiMark with an object
Currently, we need to patch `expensify-common/lib/ExpensiMark.js` to convert `class ExpensiMark` into a single object because Hermes does not support ES Classes (https://github.com/facebook/hermes/issues/685#issuecomment-1784172071).

In particular, we had to make the following changes:

1. Remove `class` keyword:
```diff
-export default class ExpensiMark {
+export const ExpensiMark = {
```

2. Rename constructor to initializer:
```diff
-    constructor() {
+    initializer() {
```

3. Add commas after method definitions:
```diff
-    }
+    },
```

It would be better to use a Babel plugin to automatically convert `class ExpensiMark` into something that Hermes can deal with.

BartoszGrajdek:
Closes #41


tomekzaw:
Closing this because #360 has been merged.

PR #149, opened 2024-02-02
Add macOS support

tomekzaw:
### Related Issues

GH_LINK


matthargett:
visionOS support also coming? :)

src/MarkdownTextInput.tsx
```diff
@@ -9,6 +9,7 @@ type MarkdownStyle = MarkdownTextInputDecoractorView.MarkdownStyle;
 
 const FONT_FAMILY_MONOSPACE = Platform.select({
   ios: 'Courier',
+  macos: 'Courier',
```
Simek, line 12:
```suggestion
  macos: 'SF Mono',
```

Can we use `SF Mono` as default for macOS? 
* https://developer.apple.com/fonts/
* https://en.wikipedia.org/wiki/San_Francisco_(sans-serif_typeface)

tomekzaw: Sure, thanks for this suggestion! Do we want to stick with `Courier` on iOS though?

Simek: Done a bit of checking, and looks like SF fonts are not included by default to this day on both iOS and macOS. In this case `Menlo` or `Monaco` could be a viable alternative:
* https://developer.apple.com/fonts/system-fonts/
* https://en.wikipedia.org/wiki/Menlo_(typeface)
* https://en.wikipedia.org/wiki/Monaco_(typeface)

hmbrg:
Any updates on this one? MacOS support would be huge!

Saadnajmi:
Hello! I would be interested if this PR was revived, though as I understand it, now this library is new architecture only?
Could I get an idea of any blockers y'all were seeing in macOS if there were any?

tomekzaw:
@Saadnajmi Hey, nice to hear that!

Yep, as of today `@expensify/react-native-live-markdown` supports only the new architecture, see explanation in https://github.com/Expensify/react-native-live-markdown/pull/634.

So one of the blockers is that we're still using method swizzling to override the default behavior when typing or updating the value from JS. There's a PR that eliminates method swizzling but there are still some details that need to be fixed: https://github.com/Expensify/react-native-live-markdown/pull/520

As far as I remember, there were no blockers caused by `react-native-macos` so far, only some API differences like lack of `UITextRange` (we'll need to use `NSRange` instead) or some problems with `valueForKey:`.

I'm curious to know about your use case, feel free to send me a DM.

Saadnajmi:
> @Saadnajmi Hey, nice to hear that!
> 
> 
> 
> Yep, as of today `@expensify/react-native-live-markdown` supports only the new architecture, see explanation in https://github.com/Expensify/react-native-live-markdown/pull/634.
> 
> 
> 
> So one of the blockers is that we're still using method swizzling to override the default behavior when typing or updating the value from JS. There's a PR that eliminates method swizzling but there are still some details that need to be fixed: https://github.com/Expensify/react-native-live-markdown/pull/520
> 
> 
> 
> As far as I remember, there were no blockers caused by `react-native-macos` so far, only some API differences like lack of `UITextRange` (we'll need to use `NSRange` instead) or some problems with `valueForKey:`.
> 
> 
> 
> I'm curious to know about your use case, feel free to send me a DM.

The use case is mostly "it would be real great to have a proper open source richtextinput library for desktop". It feels like a common request for desktop apps specifically. We have some custom stuff internally, but it's not something that would be useful in open source IMO.

apple/RCTMarkdownStyle.h
```diff
@@ -1,34 +1,34 @@
-#import <UIKit/UIKit.h>
+#import <React/RCTUIKit.h>
```
Saadnajmi, line 1: I’ve put uses of RCTUIKit inside an ifdef for other libraries, because I don’t think iOS would compile since React Native Core doesn’t have it

apple/RCTMarkdownStyle.h
```diff
@@ -1,34 +1,34 @@
...
 
 #import <react/renderer/components/RNLiveMarkdownSpec/Props.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
 @interface RCTMarkdownStyle : NSObject
 
-@property (nonatomic) UIColor *syntaxColor;
```
Saadnajmi, line 9: I think React Native macOS #defines UIColor to NSColor if that helps?

apple/RCTMarkdownStyle.h
tomekzaw: I'll add `#ifdef`s later on, now I'm just trying to make it compile on macOS.

PR #150, opened 2024-02-05
Fix missing headers on iOS when using static linkage

tomekzaw:
### Details

This PR fixes the following compile errors related to missing headers on iOS when static linkage is enabled (reported by @staszekscp):

Screenshot 2024-02-05 at 13 33 40

As it turns out, when static linkage is enabled, all `-` in header paths are replaced with `_`. The following change resolves the problem:
```diff
-#import <react-native-live-markdown/RCTMarkdownUtils.h>
+#import <react_native_live_markdown/RCTMarkdownUtils.h>
```

However, we would need to add macro conditionals to support both paths like Expo does here:

https://github.com/expo/expo/blob/294399de1e40f628b3ffdd5176eb181690d870ce/packages/expo-dev-menu/ios/ReactNativeCompatibles/ReactNative/DevClientAppDelegate.mm#L5-L10

The simpler solution is to eliminate all `-` from podspec name by renaming it from `react-native-live-markdown` to `RNLiveMarkdown` (similarly to `RNReanimated`, `RNGestureHandler` and others).

### Related Issues

GH_LINK

### Manual Tests

```
cd ios && USE_FRAMEWORKS=static NO_FLIPPER=1 pod install
```

### DIFF
--- a/ios/MarkdownTextInputDecoratorView.mm
+++ b/ios/MarkdownTextInputDecoratorView.mm
@@ -1,15 +1,15 @@
 #import <React/RCTUITextField.h>
 #import "react_native_assert.h"
 
-#import <react_native_live_markdown/MarkdownLayoutManager.h>
-#import <react_native_live_markdown/MarkdownTextInputDecoratorView.h>
-#import <react_native_live_markdown/RCTBackedTextFieldDelegateAdapter+Markdown.h>
-#import <react_native_live_markdown/RCTUITextView+Markdown.h>
+#import <RNLiveMarkdown/MarkdownLayoutManager.h>
+#import <RNLiveMarkdown/MarkdownTextInputDecoratorView.h>
+#import <RNLiveMarkdown/RCTBackedTextFieldDelegateAdapter+Markdown.h>
+#import <RNLiveMarkdown/RCTUITextView+Markdown.h>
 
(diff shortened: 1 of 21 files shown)


ios/MarkdownTextInputDecoratorView.mm
```diff
@@ -1,15 +1,15 @@
 #import <React/RCTUITextField.h>
-#import <react/debug/react_native_assert.h>
+#import "react_native_assert.h"
```
tomekzaw, line 2: Screenshot 2024-02-05 at 13 56 33

ios/RCTMarkdownStyle.mm
```diff
@@ -1,9 +1,9 @@
-#import <react-native-live-markdown/RCTMarkdownStyle.h>
+#import <RNLiveMarkdown/RCTMarkdownStyle.h>
 
 #ifdef RCT_NEW_ARCH_ENABLED
 #import <React/RCTConversions.h>
 #else
-#import <React-Core/React/RCTConvert.h>
+#import <React/RCTConvert.h>
```
tomekzaw, line 6: `React_Core` doesn't work

AndrewGable (approved):

robertKozik (approved):

thienlnam (approved):

PR #153, opened 2024-02-05
Change npm usage to yarn inside publish.yml

robertKozik:
### Details
This PR migrates publish.yml github action from using `npm` to `yarn`

### Related Issues

GH_LINK


tomekzaw (approved):

AndrewGable (dismissed):
I think we'd prefer to stick to npm, what is stoping us from doing so?

thienlnam (approved):
Discussion here - keeping as is for now

https://expensify.slack.com/archives/C06BDSWLDPB/p1707164028418069

PR #200, opened 2024-02-23
Pass ranges as objects instead of arrays

robertKozik:
### Related Issues

GH_LINK


tomekzaw (approved):
Tested on Android, iOS and web.

PR #209, opened 2024-03-01
Diff-based text replacing

BartoszGrajdek:
### Details

This PR fixes a couple of problems mainly:
Text replacement issue on MacOS (chrome & safari, not supported on Firefox) 
Spell-check on Chrome & safari (another PR will solve this issue on Firefox)

### Related Issues

https://github.com/Expensify/App/issues/36188
https://github.com/Expensify/App/issues/36193

### Manual Tests

For text replacement:
1. Make sure that you have a text replacement available on your mac more info [here](https://support.apple.com/en-am/guide/mac-help/mh35735/mac) - I have ,,cu" set up to be replaced with "see you"
2. In your text input add some text like ,,hello world"
3. Now try to add the text replacement i.e. I tried writing "current", in the bug mentioned it was replaced by "see yourrent", check if that's not happening anymore

For spellcheck:
1. Try writing a word like "apple", while you're writing it shouldn't be underlined with a red curly line
2. Now write an incorrect word like "appl" - depending on your browser spellcheck happens at different times, but overall it shouldn't happen while you're continuously writing a word

### DIFF
--- a/src/web/parserUtils.ts
+++ b/src/web/parserUtils.ts
@@ -185,9 +185,13 @@ function parseText(
 
     const rootSpan = targetElement.firstChild as HTMLElement | null;
     if (!rootSpan || rootSpan.innerHTML !== dom.innerHTML) {
-      targetElement.innerHTML = '';
-      targetElement.innerText = '';
-      target.appendChild(dom);
+      if (rootSpan) {
+        rootSpan.replaceWith(dom);
+      } else {
+        targetElement.innerHTML = '';
+        targetElement.innerText = '';
+        target.appendChild(dom);
+      }
 
       if (alwaysMoveCursorToTheEnd) {
         CursorUtils.moveCursorToEnd(target);
--- a/src/web/parserUtils.ts
+++ b/src/web/parserUtils.ts
@@ -172,13 +172,10 @@ function moveCursor(isFocused: boolean, alwaysMoveCursorToTheEnd: boolean, curso
   }
 }
 
-type Window = {
-  chrome: unknown;
-};
+const isChromium = 'chrome' in window;
 
 function parseText(target: HTMLElement, text: string, curosrPositionIndex: number | null, markdownStyle: PartialMarkdownStyle = {}, alwaysMoveCursorToTheEnd = false) {
   const targetElement = target;
-  const isChromium = (window as unknown as Window).chrome;
 
   let cursorPosition: number | null = curosrPositionIndex;
   const isFocused = document.activeElement === target;
--- a/src/web/parserUtils.ts
+++ b/src/web/parserUtils.ts
@@ -179,8 +179,6 @@ function parseText(
 
   const markdownRanges: MarkdownRange[] = ranges as MarkdownRange[];
 
-  let appended = false;
-
   // We don't want to parse text with single '\n', because contentEditable represents it as invisible <br />
   if (!!text && text !== '\n') {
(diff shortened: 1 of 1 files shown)


BartoszGrajdek:

BartoszGrajdek:
recheck

src/web/parserUtils.ts
```diff
@@ -179,19 +179,22 @@ function parseText(
...
     const dom = parseRangesToHTMLNodes(text, markdownRanges, markdownStyle);
-    target.appendChild(dom);
-  }
 
-  if (alwaysMoveCursorToTheEnd) {
-    CursorUtils.moveCursorToEnd(target);
-  } else if (isFocused && cursorPosition !== null) {
-    CursorUtils.setCursorPosition(target, cursorPosition, disableNewLinesInCursorPositioning);
+    // eslint-disable-next-line es/no-optional-chaining
+    if ((targetElement.firstChild as HTMLElement | null)?.innerHTML !== dom.innerHTML) {
```
Skalakid, line 187: I'm not a fan of disabling eslint rules, what about changing it to something like this:
```suggestion
if (!targetElement.firstChild || (targetElement.firstChild as HTMLElement).innerHTML !== dom.innerHTML) {
```

tomekzaw: Also, could you please explain why do we need to access `.firstChild` of `targetElement` and not the `targetElement` itself?

src/web/parserUtils.ts
```diff
@@ -179,19 +179,22 @@ function parseText(
...
+      targetElement.innerHTML = '';
+      targetElement.innerText = '';
+      target.appendChild(dom);
```
tomekzaw, line 190: How about using `replaceWith` method instead? https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceWith

src/web/parserUtils.ts
BartoszGrajdek: This is because the `target` element is the div itself, so if we try retrieving it's `innerHTML` it will include root span, whereas if we retrieve `innerHTML` from `dom` it won't be including the root span @tomekzaw

src/web/parserUtils.ts
BartoszGrajdek: It won't work since there's no root span when the component is rendered for the first time, because of which there's no way to replace it

Skalakid (approved):

tomekzaw:
There seems to be a problem with the position of text replacement suggestion box in Chrome:

https://github.com/Expensify/react-native-live-markdown/assets/20516055/82ba6d76-b871-4aab-8fe8-7a2e2a47c054

The suggestion box is not visible on Safari:

https://github.com/Expensify/react-native-live-markdown/assets/20516055/6750e2e6-58d2-4a82-8621-8a140af728cb

tomekzaw:
I also found one regression on Safari, after adding newline the cursor disappears:

<table>
<thead>
<tr>
<th>main</th>
<th>this PR</th>
</tr>
</thead>
<tbody>
<tr>
<td><video src="https://github.com/Expensify/react-native-live-markdown/assets/20516055/c89b1208-6551-4cc1-99cc-0f36afbfb654"></td>
<td><video src="https://github.com/Expensify/react-native-live-markdown/assets/20516055/fb747604-d20f-45f5-8879-52b9d902902c"></td>
</tr>
</tbody>
</table>

tomekzaw (approved):

PR #210, opened 2024-03-04
Diacritics fix

BartoszGrajdek:
### Details

This PR fixes diacritics not working on every browser

### Related Issues

https://github.com/Expensify/App/issues/36942

### Manual Tests

1. Try adding a diacritic in the text input i.e. Option+u and then "a" - this should result in "ä"
2. You can now try different cases like doing it in nested blocks, at the beginning of a word, in the middle of it, or by itself.

### DIFF
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -499,13 +499,16 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
     }, []);
 
     useEffect(() => {
-      document.documentElement.style.setProperty('--placeholder-color', placeholderTextColor as string);
       // focus the input on mount if autoFocus is set
       if (!(divRef.current && autoFocus)) {
         return;
       }
       divRef.current.focus();
     }, []);
+
+    useEffect(() => {
+      document.documentElement.style.setProperty('--placeholder-color', placeholderTextColor as string);
+    }, [placeholderTextColor]);
 
     const startComposition = useCallback(() => {
       compositionRef.current = true;
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -258,6 +258,12 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
       return e;
     }, []);
 
+    // Placeholder text color logic
+    const updateTextColor = useCallback((node: HTMLDivElement, text: string) => {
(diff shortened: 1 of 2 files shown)


Skalakid (dismissed):
LGTM

src/MarkdownTextInput.web.tsx
```diff
@@ -156,6 +156,7 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
     },
     ref,
   ) => {
+    const [isComposition, setIsComposition] = useState(false);
```
tomekzaw, line 159: Looks like we never read `isComposition` during React render and we probably also don't want to trigger a React render when this changes. How about using `useRef` here instead?

src/MarkdownTextInput.web.tsx
```diff
@@ -156,7 +156,7 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
     },
     ref,
   ) => {
-    const [isComposition, setIsComposition] = useState(false);
+    const compositionRef = useRef<boolean | null>(null);
```
tomekzaw, line 159: Does it work if we initialize the ref with `false`?
```suggestion
    const compositionRef = useRef<boolean>(false);
```

src/MarkdownTextInput.web.tsx
```diff
@@ -520,6 +520,10 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
       divRef.current.focus();
     }, []);
 
+    const startComposition = () => {
+      compositionRef.current = true;
+    };
```
tomekzaw, line 525: Can we memoize this callback so it's not regenerated on each render?
```suggestion
    const startComposition = useCallback(() => {
      compositionRef.current = true;
    }, []);
```

tomekzaw:
There seems to be some problem with text color in this case:

https://github.com/Expensify/react-native-live-markdown/assets/20516055/ae66fed0-ccfc-4266-b7b2-b3a3545ace73

Skalakid (approved):

Skalakid (changes requested):
Placeholder color doesn't update when `placeholderTextColor` prop changes


https://github.com/Expensify/react-native-live-markdown/assets/39538890/ec24aa90-8678-4b2d-9f7c-00d47f07a8f0

BartoszGrajdek:
Fixed @Skalakid

Skalakid (dismissed):
LGTM

tomekzaw (approved):
Thanks for fixing this!

PR #215, opened 2024-03-08
Fix Shift + Enter on Safari

Skalakid:
### Details

This PR changes the default behavior of Shift + Enter to be the same as when clicking Enter. Thanks to that we have control over what happens after inserting new line and `contenteditable` work the same  on all browsers

### Related Issues

GH_LINK

### Manual Tests

1. Open example app on Safari
2. Delete whole text
3. Start entering new lines using Shift + Enter
4. Verify if cursor is correctly positioned


tomekzaw (approved):

BartoszGrajdek (approved):
LGTM!

PR #220, opened 2024-03-11
Fix scroll on enter

BartoszGrajdek:
### Details

This PR fix the problem with input not scrolling into view the cursor after entering the new line using `Enter` or `Shift` + `Enter`

### Related Issues

GH_LINK

### Manual Tests

1. Set static input height
2. Add some text so you can scroll text inside markdown input
3. On the end of the input start entering newlines. Verify on all browsers if input is scrolling down so the cursor in newline is fully visible


BartoszGrajdek:
Adding this as a separate issue ^

Skalakid (approved):

tomekzaw (approved):

PR #222, opened 2024-03-12
Fix selection setting when pasting text and auto-focusing

Skalakid:
### Details

This is follow up PR that fix setting cursor position when pasting blocks of text and autofocusing the markdown input after page loaded

### Related Issues

GH_LINK

### Manual Tests

1. Set `autoFocus` prop and verify if after entering/refreshing the page cursor is set to the end of the text inside the input
2. Verify if cursor position is correct after pasting text into markdown input


tomekzaw (approved):

BartoszGrajdek (approved):
LGTM!

PR #261, opened 2024-04-03
Fix useCallback dependency arrays

Skalakid:
### Details

This PR fixes `onSubmitEditing` prop that wasn't updated in `handleKeyPress` function. Also this PR adds `react-hooks` plugin to .eslintrc.js
### Related Issues

GH_LINK

### Linked PRs

https://github.com/Expensify/App/issues/39189

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,7 @@ module.exports = {
   },
   extends: [
     'expensify',
+    "plugin:react-hooks/recommended",
     'plugin:@typescript-eslint/recommended',
     'plugin:@typescript-eslint/stylistic',
     'plugin:import/typescript',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,7 @@ module.exports = {
   },
   extends: [
     'expensify',
-    "plugin:react-hooks/recommended",
+    'plugin:react-hooks/recommended',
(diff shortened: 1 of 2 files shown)


tomekzaw (dismissed):

src/MarkdownTextInput.web.tsx
```diff
@@ -522,6 +528,7 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
       if (autoFocus) {
         divRef.current.focus();
       }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
tomekzaw, line 531: Why do we ignore the deps here?

.eslintrc.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
   },
   extends: [
     'expensify',
+    "plugin:react-hooks/recommended",
```
tomekzaw, line 12:
```suggestion
    'plugin:react-hooks/recommended',
```

src/MarkdownTextInput.web.tsx
Skalakid: I've added this so it will launch only after the first render. But after reconsideration, I think we can remove it

tomekzaw (approved):

PR #266, opened 2024-04-05
Fix undoing pasted text into the input

BartoszGrajdek:
### Details

This issue fixes a situation where the initial value of the input doesn't get saved into `InputHistory`, because of this it's not possible to undo the previous action

### Related Issues

[GH_LINK](https://github.com/Expensify/App/issues/39255)

### Manual Tests

To test this remove the default value provided for our example.
1. Copy some longer text
2. Paste this text into input
3. Try to undo it with `CMD + Z`

### DIFF
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -546,6 +546,13 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
       updateSelection(null, {start: selection.start, end: selection.end || selection.start});
     }, [selection, updateSelection]);
 
+    useEffect(() => {
+      if (history.current?.history.length !== 0) {
+        return;
+      }
+      history.current.add(value ?? '', (value ?? '').length);
+    }, []);
+
     return (
       // eslint-disable-next-line jsx-a11y/no-static-element-interactions
       <div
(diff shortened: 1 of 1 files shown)


src/MarkdownTextInput.web.tsx
```diff
@@ -546,6 +546,14 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
       updateSelection(null, {start: selection.start, end: selection.end || selection.start});
     }, [selection, updateSelection]);
 
+    useEffect(() => {
+      if (history.current?.history.length !== 0) {
+        return;
+      }
+      history.current.add(value ?? '', (value ?? '').length);
```
Skalakid, line 553:
```suggestion
      if (!value || history.current?.history.length !== 0) {
        return;
      }
      history.current.add(value, value.length);
```

BartoszGrajdek: This wouldn't work, I actually do want the history item to be added if value is `undefined`, or just an empty string - that's what we're fixing here

Skalakid: Right, so maybe let's move `value ?? ''` to variable since we are using it two times

Skalakid (approved):

PR #279, opened 2024-04-09
Improve error handling in `parseExpensiMarkToRanges`

BartoszGrajdek:
### Details

This PR changes how we handle errors in `parseExpensiMarkToRanges`. Examples should now ignore styling if they encounter error while parsing instead of crashing. 

It also improves the error messages itself so that they explain what is failing, where and why 

### Related Issues

GH_LINK

### Manual Tests

Add a `new Error()` somewhere in the file or just make sure one of the existing errors gets thrown


parser/index.ts
```diff
@@ -227,17 +227,23 @@ function groupRanges(ranges: Range[]) {
...
-  if (text !== markdown) {
-    // text mismatch, don't return any ranges
+  try {
+    const html = parseMarkdownToHTML(markdown);
+    const tokens = parseHTMLToTokens(html);
+    const tree = parseTokensToTree(tokens);
+    const [text, ranges] = parseTreeToTextAndRanges(tree);
+    if (text !== markdown) {
+      // text mismatch, don't return any ranges
+      return [];
```
robertKozik, line 237: We can throw an error here, so we will return empty array only in catch block

parser/index.ts
```diff
@@ -227,17 +238,24 @@ function groupRanges(ranges: Range[]) {
...
-    // text mismatch, don't return any ranges
+  try {
+    const html = parseMarkdownToHTML(markdown);
+    const tokens = parseHTMLToTokens(html);
+    const tree = parseTokensToTree(tokens);
+    const [text, ranges] = parseTreeToTextAndRanges(tree);
+    if (text !== markdown) {
+      throw new Error(
+        `[react-native-live-markdown] Parsing error: the processed text does not match the original markdown input. This may be caused by incorrect parsing functions or invalid input markdown.`,
+      );
```
robertKozik, line 249: Can we add the values of text and markdown to the error message
```suggestion
      throw new Error(
        `[react-native-live-markdown] Parsing error: the processed text does not match the original markdown input. This may be caused by incorrect parsing functions or invalid input markdown.`,
        {processed: text, original: markdown}
      );
```

BartoszGrajdek: Unfortunately, this is not how this works, we would need to create a CustomError extending Error to make it work. 

I picked an easier solution by directly injecting these strings into the error message

parser/index.ts
```diff
@@ -38,7 +38,7 @@ function parseHTMLToTokens(html: string): Token[] {
     }
     const close = html.indexOf('>', open);
     if (close === -1) {
-      throw new Error('Invalid HTML: no matching ">"');
+      throw new Error('[react-native-live-markdown] Invalid HTML: no matching ">"');
```
tomekzaw, line 41: Let's add `Error in function parseHTMLToTokens:` for consistency with other error messages.

parser/index.ts
```diff
@@ -227,17 +238,26 @@ function groupRanges(ranges: Range[]) {
...
+  try {
+    const html = parseMarkdownToHTML(markdown);
+    const tokens = parseHTMLToTokens(html);
+    const tree = parseTokensToTree(tokens);
+    const [text, ranges] = parseTreeToTextAndRanges(tree);
+    if (text !== markdown) {
+      throw new Error(
+        `[react-native-live-markdown] Parsing error: the processed text does not match the original markdown input. This may be caused by incorrect parsing functions or invalid input markdown.\nProcessed text: '${JSON.stringify(
+          text,
+        )}'\nOriginal markdown: '${JSON.stringify(markdown)}'`,
```
tomekzaw, line 250:
```suggestion
        `[react-native-live-markdown] Parsing error: the processed text does not match the original Markdown input. This may be caused by incorrect parsing functions or invalid input Markdown.\nProcessed input: '${JSON.stringify(
          text,
        )}'\nOriginal input: '${JSON.stringify(markdown)}'`,
```

tomekzaw (approved):
Thanks for this PR!

Skalakid (approved):

robertKozik (approved):

PR #283, opened 2024-04-11
fix: case element height equal 0

nkdengineer:
### Details

Currently, we didn't check if the height is positive before returning the value. The height could be 0 if the parent element was not fully rendered in the DOM yet.

So on layout change, when we use the method to get the element height [here](https://github.com/Expensify/react-native-live-markdown/blob/c7595572ff4f2a4b8c95a6110a05de62b2d8dbe4/src/MarkdownTextInput.web.tsx#L524) to use as maxHeight, we're getting 0 as the height

To fix this:  if the height is 0, we should return auto as the height

This is similar to the logic when numberOfLines is not defined [here](https://github.com/Expensify/react-native-live-markdown/blob/c7595572ff4f2a4b8c95a6110a05de62b2d8dbe4/src/MarkdownTextInput.web.tsx#L123) . There's a bug there also, the condition after || will never be triggered, it should be `styles.height ? '${styles.height}px' : 'auto'.`

### Related Issues

GH_LINK

### DIFF
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -117,10 +117,10 @@ function getElementHeight(node: HTMLDivElement, styles: CSSProperties, numberOfL
       node.parentElement.appendChild(tempElement);
       const height = tempElement.clientHeight;
       node.parentElement.removeChild(tempElement);
-      return `${height}px`;
+      return height ? `${height}px` : 'auto';
     }
   }
-  return `${styles.height}px` || 'auto';
+  return styles.height ? `${styles.height}px` : 'auto';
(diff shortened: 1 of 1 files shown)


tomekzaw:
@nkdengineer Thanks for the PR! We'll review it shortly

nkdengineer:

BartoszGrajdek:
Reviewing! 👀

BartoszGrajdek (approved):
LGTM! 🚀 

cc @tomekzaw please merge this once you think we can safely upgrade the version of `react-native-live-markdown` in Expensify 🙏🏻

PR #284, opened 2024-04-11
Filter built ABIs so the library compiles when building for select ABIs

j-piasecki:
### Details
The library is always being built for all available ABIs even when the app it's being used in is only built for one. This causes it to fail on linking stage if React Native is built from source, since relevant `.so` files from React Native core don't exist. This PR adds relevant filters.

### Manual Tests
Run `yarn build:android` from the example app.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,6 +9,11 @@ buildscript {
   }
 }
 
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):
It was a tough one 😄

PR #288, opened 2024-04-11
Restore backed text view in MarkdownUtils

j-piasecki:
### Details
Restores setting backed text view in MarkdownUtils which was removed in https://github.com/Expensify/react-native-live-markdown/pull/199

### Manual Tests
I'll be honest, I have no idea where to look for the effects of this change. I don't see a difference in codeblocks after changing it back, so it may simply not work or I don't know what to look for.


tomekzaw:
Closing this PR in favor of #289.

PR #299, opened 2024-04-12
Remove `pod-install` script

tomekzaw:
### Details

This PR gets rid of `pod-install` script that automatically installed Pods for iOS when running yarn to install node_modules.

The reason is that `pod install` is generally slow and we don't want to run it each time we re-install node_modules.

Also, the recommended usage is `bundler install && bundler exec pod install`.

### Related Issues

GH_LINK


BartoszGrajdek (approved):
Makes sense to me! 👀

PR #300, opened 2024-04-12
Fix iOS build cache reuse

tomekzaw:
### Details
This PR changes the cache key used for restoring iOS dependencies (Pods) and build artifacts (DerivedData).

Currently, the cache key includes a hash calculated from `example/ios/Podfile.lock`. Unfortunately, this file changes very frequently as we automatically bump patch version after each merged PR which results in updating the caches on each PR which is time-consuming.

Most likely, the only scenario where we should drop the caches is when we bump `react-native` version in our example app and that's exactly what this PR does.

### Related Issues

GH_LINK

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -52,7 +52,7 @@ jobs:
             example/ios/Pods
             ~/Library/Caches/CocoaPods
             ~/.cocoapods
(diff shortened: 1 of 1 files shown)


BartoszGrajdek (approved):
👍🏻

PR #306, opened 2024-04-12
Implement code and pre blocks support on iOS

maksg:
### Details
Adds support for displaying and customising code and pre blocks on iOS.


tomekzaw:
Still discussing the expected behavior for codeblocks [here](https://swmansion.slack.com/archives/C06BDSWLDPB/p1713274778358479)

tomekzaw:
It looks like `padding` in `markdownStyle` only affects the size of the rectangle in the background and not the text itself:

Screenshot 2024-04-17 at 01 35 09

It's okay for the initial implementation but ultimately we'd like `padding` to affect the text layout here as well.

I don't think it's possible to achieve padding for inline code on iOS (Slack for iOS also does not support it).

However, I think it should be possible to implement padding for codeblocks. As for horizontal padding, I've copied this code from blockquotes and it seems to work:

https://github.com/Expensify/react-native-live-markdown/blob/cd42c45c0a76c4beba86e52899e196d9f176eb46/ios/RCTMarkdownUtils.mm#L119-L122

Screenshot 2024-04-17 at 01 36 38

As for vertical padding, ChatGPT says that the only option is to set `lineSpacing` but this affects the height of the cursor so I'm not really sure how to implement it.

tomekzaw:
There seems to be some problem with codeblocks containing only one character:

https://github.com/Expensify/react-native-live-markdown/assets/20516055/935e409f-8dae-4829-8a51-9d877e905534

tomekzaw:
Closing in favor of https://github.com/Expensify/react-native-live-markdown/pull/720. Thanks again for the hard work!

PR #319, opened 2024-04-17
Fix input focusing on selection prop change

Skalakid:
### Related Issues

[GH_LINK](https://github.com/Expensify/App/issues/36071#issuecomment-2031031366)

### DIFF
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -437,14 +437,12 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
         currentlyFocusedField.current = hostNode;
         setEventProps(e);
         if (divRef.current) {
-          const valueLength = value ? value.length : 0;
-          let startPosition = valueLength;
-          let endPosition = null;
           if (contentSelection.current) {
-            startPosition = contentSelection.current.start;
-            endPosition = contentSelection.current.end;
+            CursorUtils.setCursorPosition(divRef.current, contentSelection.current.start, contentSelection.current.end);
+          } else {
+            const valueLength = value ? value.length : divRef.current.innerText.length;
+            CursorUtils.setCursorPosition(divRef.current, valueLength, null);
           }
-          CursorUtils.setCursorPosition(divRef.current, startPosition, endPosition);
           updateSelection(event, contentSelection.current);
         }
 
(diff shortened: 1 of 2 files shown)


src/web/cursorUtils.ts
```diff
@@ -28,6 +28,9 @@ function setPrevText(target: HTMLElement) {
 }
 
 function setCursorPosition(target: HTMLElement, start: number, end: number | null = null) {
+  if (target !== document.activeElement) {
+    return;
```
tomekzaw, line 32: Maybe let's add a comment why we need to return here?

src/MarkdownTextInput.web.tsx
```diff
@@ -438,8 +438,14 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
...
           const valueLength = value ? value.length : 0;
-          CursorUtils.setCursorPosition(divRef.current, contentSelection.current ? contentSelection.current.end : valueLength);
-          updateSelection(event);
+          let startPosition = valueLength;
+          let endPosition = null;
+          if (contentSelection.current) {
+            startPosition = contentSelection.current.start;
+            endPosition = contentSelection.current.end;
+          }
+          CursorUtils.setCursorPosition(divRef.current, startPosition, endPosition);
```
tomekzaw, line 447: I don't like that we create two non-const variables here.

Suggestion 1:
```suggestion
          if (contentSelection.current) {
            CursorUtils.setCursorPosition(divRef.current, contentSelection.current.start, contentSelection.current.end);
          } else {
            CursorUtils.setCursorPosition(divRef.current, valueLength, null);
          }
```

Suggestion 2:
```suggestion
          const [startPosition, endPosition] = contentSelection.current
            ? [contentSelection.current.start, contentSelection.current.end]
            : [valueLength, null];
          CursorUtils.setCursorPosition(divRef.current, startPosition, endPosition);
```

Suggestion 3:
```suggestion
          CursorUtils.setCursorPosition(
            divRef.current,
            contentSelection.current ? contentSelection.current.start : valueLength,
            contentSelection.current ? contentSelection.current.end : null);
```

tomekzaw (changes requested):

src/MarkdownTextInput.web.tsx
Skalakid: I like suggestion number 1, thanks

PR #320, opened 2024-04-18
Ensure caret is visible

ntdiary:
### Details

https://github.com/Expensify/react-native-live-markdown/blob/cd42c45c0a76c4beba86e52899e196d9f176eb46/src/MarkdownTextInput.web.tsx#L441

In Chrome and Safari (mobile & PC), for long text message, even though we set the cursor position when focused, `maxHeight` hasn't taken effect yet, so the element has no overflow, causing the caret to be invisible after rendering. This change ensures that the caret is visible.

https://github.com/Expensify/react-native-live-markdown/assets/8579651/e13cf405-39a8-44ce-ba70-11516f0b4b29

### Related Issues

https://github.com/Expensify/App/issues/26239#issuecomment-2059414561

### Linked PRs

https://github.com/Expensify/App/pull/40295

### DIFF
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -553,6 +553,7 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
         const elementHeight = getElementHeight(divRef.current, inputStyles, numberOfLines);
         divRef.current.style.height = elementHeight;
         divRef.current.style.maxHeight = elementHeight;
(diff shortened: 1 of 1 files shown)


Skalakid:
Hi @ntdiary, thanks for the PR. I'm not sure if we want to scroll to the cursor every time `numberOfLines` prop changes. It can introduce some other regressions in the future. Also, we believe it more of a bug on the Expensify App side.
However, recently I've been working on enhancing the Composer component, and witch my changes tour problem is fixed. here is [PR](https://github.com/Expensify/App/pull/40128)

PR #360, opened 2024-05-30
bump expensify-common version

nkdengineer:
### Details

Bump new expensify-common version to fix the phone number mention for live-markdown preview
### Related Issues

https://github.com/Expensify/App/issues/41078

### Linked PRs

https://github.com/Expensify/expensify-common/pull/697

### DIFF
--- a/parser/package.json
+++ b/parser/package.json
@@ -20,7 +20,7 @@
     "typescript": "^5.3.3"
   },
   "dependencies": {
-    "expensify-common": "Expensify/expensify-common#a018512b1e0fce4d4b5a61c00ca826b8887007ad",
+    "expensify-common": "Expensify/expensify-common#a451327f6c0f516463d9da7c6aaa1c1c63b1a960",
     "patch-package": "^8.0.0",
     "underscore": "^1.13.6"
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2388,7 +2388,7 @@ __metadata:
     "@types/underscore": ^1.11.15
     esbuild: 0.19.4
     esbuild-plugin-tsc: ^0.4.0
-    expensify-common: "Expensify/expensify-common#a018512b1e0fce4d4b5a61c00ca826b8887007ad"
+    expensify-common: "Expensify/expensify-common#a451327f6c0f516463d9da7c6aaa1c1c63b1a960"
     jest: ^29.7.0
     patch-package: ^8.0.0
     typescript: ^5.3.3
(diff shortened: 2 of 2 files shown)


parser/index.ts
```diff
@@ -1,6 +1,5 @@
-// @ts-expect-error - to review how it's implemented in ExpensiMark
 // eslint-disable-next-line import/no-unresolved
-import {ExpensiMark} from 'expensify-common/lib/ExpensiMark';
+import {ExpensiMark} from 'expensify-common';
```
hungvu193, line 2:
```suggestion
import ExpensiMark from 'expensify-common/lib/ExpensiMark';
```

parser/patches/expensify-common+2.0.4.patch
hungvu193, line 1: This patch is exactly the same with `expensify-common+1.0.0.patch` right?

nkdengineer: The only different is not export ExpensiMark as `const`

parser/index.ts
nkdengineer: This import now isn't correct when install `Expensify-common` by version.

parser/index.ts
```diff
@@ -14,7 +13,7 @@ type Token = ['TEXT' | 'HTML', string];
 type StackItem = {tag: string; children: Array<StackItem | string>};
 
 function parseMarkdownToHTML(markdown: string): string {
-  const parser = ExpensiMark;
+  const parser = new ExpensiMark();
```
tomekzaw, line 16: I don't think it will work as Hermes doesn't still support classes and we don't transpile this code when generating Live Markdown JS bundle.
```suggestion
  const parser = ExpensiMark;
```

nkdengineer: @tomekzaw I updated the patch to export `ExpensiMark` as const.

parser/package.json
```diff
@@ -20,7 +20,7 @@
     "typescript": "^5.3.3"
   },
   "dependencies": {
-    "expensify-common": "Expensify/expensify-common#a018512b1e0fce4d4b5a61c00ca826b8887007ad",
+    "expensify-common": "2.0.4",
```
tomekzaw, line 23: Can we update to the latest (current latest is 2.0.9)?

blimpich:
@hungvu193 is this PR ready for internal engineer review other than the version bump specified [here](https://github.com/Expensify/react-native-live-markdown/pull/360#discussion_r1627867723)? Lets keep this PR moving 🏃‍♂️‍➡️ 🙂

hungvu193:
> @hungvu193 is this PR ready for internal engineer review other than the version bump specified [here](https://github.com/Expensify/react-native-live-markdown/pull/360#discussion_r1627867723)? Lets keep this PR moving 🏃‍♂️‍➡️ 🙂

All good. Just a comment left that you mentioned 😄

parser/package.json
nkdengineer: @tomekzaw I just updated to the latest version. Please help to check the patch again, I see that we have a logger on the latest version.

tomekzaw: Thanks for bumping, will review it now.

parser/index.ts
tomekzaw: Fixed in 1cfd1eb.

tomekzaw: Actually I had to revert this change as it caused the whole `expensify-common` to bundle and thus significantly increase its bundle size. I guess it's okay to leave it as it is now since the types also work correctly.

tomekzaw:
I was able to eliminate `expensify-common` patch as well as remove `patch-package` from `parser` directory by using Babel (in particular, `module:@react-native/babel-preset`) to transpile the bundled file.

tomekzaw (dismissed):

tomekzaw (approved):

PR #363, opened 2024-06-01
Fix: correct content height in TextKit 1 compatibility mode

QichenZhu:
### Details

The input is sometimes not scrollable on iOS.

### Related Issues

$ https://github.com/Expensify/App/issues/41567
PROPOSAL: https://github.com/Expensify/App/issues/41567#issuecomment-2137792675

### Manual Tests

1. Open New Expensify and log in
2. Tap FAB > Assign task
3. Enter any title and add a long comment
4. On the confirmation screen open the the description and scroll down
5. Finish creating a task
6. Open the created task and tap the description
7. Scroll down

Expected Result:
User is able to scroll down the task description.

### DIFF
--- a/ios/MarkdownTextInputDecoratorView.mm
+++ b/ios/MarkdownTextInputDecoratorView.mm
@@ -72,6 +72,14 @@
     _textView = (RCTUITextView *)_backedTextInputView;
     [_textView setMarkdownUtils:_markdownUtils];
     NSLayoutManager *layoutManager = _textView.layoutManager; // switching to TextKit 1 compatibility mode
+
+    // Correct content height in TextKit 1 compatibility mode. (See https://github.com/Expensify/App/issues/41567)
+    // Consider removing this fix if it is no longer needed after migrating to TextKit 2.
+    CGSize contentSize = _textView.contentSize;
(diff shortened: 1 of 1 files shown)


QichenZhu:

QichenZhu:
recheck

tomekzaw:
@QichenZhu Thanks for the PR!

I might be slightly biased by the length of the code changed but I lean more towards the other solution mentioned in the comment (i.e. switching to TextKit 1 mode earlier by getting `.layoutManager` in `willMoveToWindow:`). This one is a bit shorter and thus easier to understand for me. Are there any other benefits in the current approach?

QichenZhu:
Thanks for your feedback!

I prefer the main solution because it's straightforward - the height changes after the switch, so I just recover it.

The alternative seems somewhat magical. Why does the timing of the switch matter?

Please correct me if I'm wrong. @tomekzaw @hungvu193

hungvu193:
> I prefer the main solution because it's straightforward - the height changes after the switch, so I just recover it.

I also agree with @QichenZhu at this point.

tomekzaw (approved):
I confirm this PR resolves the original issue (which btw is also reproducible in the example app). Thanks for the PR!

PR #374, opened 2024-06-07
Fix showing wrongly formatted text

zfurtak:
### Details
When typing text with `command + b` or `command + i` or `command + u` effect was applied to inputted text, which was unwanted.
Also connected issue: when text was pasted to an empty filed, it was in the colour of placeholder.
In these cases instead of `<span>` object with text, `<b>, <i>, <u>....` was being created. 

My proposal is to change the created object when it's not `<span>`


BartoszGrajdek (commented):
I think it may be easier to just change the condition in `src/web/parserUtils.ts` (L206) to check if the firstChild of our input has the class `root`. If not it should regenerate the input content, meaning creating the span & parsing the content with ExpensiMark - this will remove any additional styles.
```
if (!rootSpan || !rootSpan.classList.contains('root') || rootSpan.innerHTML !== dom.innerHTML)

zfurtak:
@BartoszGrajdek I agree this solution is much more clear but it leaves unwanted behaviour connected to placeholder.
Look below how it looks in the example app and Expensify App

https://github.com/Expensify/react-native-live-markdown/assets/92310164/833107c8-0e04-4c40-900a-9d1b46cb1039


https://github.com/Expensify/react-native-live-markdown/assets/92310164/e64f5efd-1d74-48fb-a74b-43a9658d88c0

Skalakid (approved):
LGTM

BartoszGrajdek:
We'll merge it after a version bump in E/App. To make sure in case of any regressions we don't have to add testing steps for this PR as well

PR #370, opened 2024-06-07
fix: update contentSelection ref after text change

dominictb:
### Details

https://github.com/Expensify/App/issues/41137#issuecomment-2153777091

### Related Issues

https://github.com/Expensify/App/issues/41137

### Manual Tests

Tested on Firefox

https://github.com/Expensify/react-native-live-markdown/assets/165644294/309c91fe-c8e6-472f-92ef-85375a233da5

### Linked PRs

https://github.com/Expensify/App/pull/42622

PR #357, opened 2024-05-27
fix: use setBaseAndExtent instead of removeAllRanges

dominictb:
### Details

Use `selection.setBaseAndExtent` instead of `selection.removeAllRanges(); selection.addRange(range)` to avoid the keyboard on Android chrome to hide

### Related Issues

$ https://github.com/Expensify/App/issues/41137
PROPOSAL: https://github.com/Expensify/App/issues/41137#issuecomment-2121656214

### Manual Tests

Use Android chrome and go to the main page https://127.0.0.1:8082
Tap on a report
Paste any text repeatedly and note the page

Expected: The keyboard should remain open and visible after the paste. Also, the caret position must be maintained.

### Linked PRs

https://github.com/Expensify/App/pull/42622

PR #376, opened 2024-06-09
fix: revert the functionality to restore cursor in value update hook

dominictb:
### Details

- revert the use of contentSelection to restore cursor pos during text update to cursor jumping around funny since contentSelection ref value can get outdated.
- create func to restore cursor in the ref object

### Related Issues

RCA: https://github.com/Expensify/App/issues/43312#issuecomment-2156740534
Solution: https://github.com/Expensify/App/issues/41137#issuecomment-2156739474

### DIFF
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -535,7 +535,7 @@ const MarkdownTextInput = React.forwardRef<TextInput & WithRestoreSelectionAbili
           updateTextColor(r, '');
         };
 
-        (r as any).restoreSelectionPosition = () => {
+        (r as unknown as WithRestoreSelectionAbility).restoreSelectionPosition = () => {
           if (contentSelection.current) {
             CursorUtils.setCursorPosition(r, contentSelection.current.start, contentSelection.current.end);
           } else {
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -134,11 +134,7 @@ function getElementHeight(node: HTMLDivElement, styles: CSSProperties, numberOfL
   return styles.height ? `${styles.height}px` : 'auto';
 }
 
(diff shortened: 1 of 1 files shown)


Skalakid (commented):
@dominictb I see this PR introduces some custom API functions on the web implementation. The idea behind the Live Markdown Input is to make its interface nearly identical to the native `TextInput` component so people can just replace it. Adding custom API need deeper investigation and planning. If we decide to make something new it would be great to add it on iOS and Android too.
Thank you for your changes, however I think we should try solving the issue with a bit different logic. Have you tried using the currently available features of Live Markdown to fix it on the E/App side? If yes, I think you can try fixing this issue by just changing `handlePaste` event logic in `MarkdownTextInput.web.tsx`

BartoszGrajdek:
Alright so here's what we're going to do now: We'll revert [this PR](https://github.com/Expensify/react-native-live-markdown/pull/357) since [a fix](https://github.com/Expensify/react-native-live-markdown/pull/370) for the [regression](https://github.com/Expensify/App/issues/41137#issuecomment-2153464605) I've mentioned was reverted [here](https://github.com/Expensify/react-native-live-markdown/pull/377). You'll have to make 1 PR with all of these changes and we will do our best to help you along the way with the implementation since it looks like this is a tricky bug. Our goal is to keep `react-native-live-markdown` lib as bug-free as we can and that way we can properly test it to ensure nothing breaks. 🙌🏻

Skalakid (approved):
LGTM, thanks for investigation. We will close @BartoszGrajdek's [PR](https://github.com/Expensify/react-native-live-markdown/pull/380) with the revert

BartoszGrajdek (approved):
We're basically restoring how it worked before except for `setBaseAndExtent` so it looks solid to me. I also like the idea of handling the bug in E/App itself

PR #386, opened 2024-06-17
Remove underscore from markdown parser

Skalakid:
### Details

This PR removes `underscore` library from markdown parser

### Related Issues

GH_LINK

### Manual Tests

1. Check if tests pass after these changes


.eslintrc.js
```diff
@@ -58,6 +58,7 @@ module.exports = {
       'error',
       { fixMixedExportsWithInlineTypeSpecifier: false },
     ],
+    'valid-jsdoc': 'off',
```
tomekzaw, line 61: What was the reason to disable this rule?

parser/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+const reEscapedHtml = /&(?:amp|lt|gt|quot|#(x27|x60|32));/g;
+const reHasEscapedHtml = RegExp(reEscapedHtml.source);
+
+/**
+ * The inverse of `escape`this method converts the HTML entities
+ * `&amp;`, `&lt;`, `&gt;`, `&quot;` and `&#39;` in `string` to
+ * their corresponding characters.
+ * Source: https://github.com/lodash/lodash/blob/main/src/unescape.ts
+ */
+function unescape(string: string): string {
```
tomekzaw, line 20: There's already a function called `unescape` in JS ([link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape)), so let's rename it to `unescapeText` to avoid confusion.

parser/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  '&#x27;': "'",
+  '&#x60;': '`',
+  '&#32': ' ',
+};
+
+const reEscapedHtml = /&(?:amp|lt|gt|quot|#(x27|x60|32));/g;
+const reHasEscapedHtml = RegExp(reEscapedHtml.source);
+
+/**
+ * The inverse of `escape`this method converts the HTML entities
```
tomekzaw, line 15: Something wrong with this description

.eslintrc.js
Skalakid: `valid-jsdoc` rule shows error when `@return {type} ...` or `@param {type} ...` isn't added. However when you include it `tsdoc/syntax` rule shows error because you included the type. These rules are excluding each other

tomekzaw (approved):

PR #409, opened 2024-07-02
Fix "Out of bounds" crash on iOS

tomekzaw:
### Details

This PR fixes the following crash on iOS
```
NSMutableRLEArray objectAtIndex:effectiveRange:: Out of bounds
```
when returning improper ranges, in particular:

1. when passing a range with zero length, e.g. `{ start: 10, length: 0, type: 'code' }`

Screenshot 2024-07-02 at 10 18 43

2. when passing a range that is longer than the string itself, e.g. `{ start: 0, length: 9999, type: 'bold' }`

Screenshot 2024-07-02 at 10 19 26

### Related Issues

GH_LINK

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -1053,13 +1053,13 @@ PODS:
     - React-jsi (= 0.73.4)
     - React-logger (= 0.73.4)
     - React-perflogger (= 0.73.4)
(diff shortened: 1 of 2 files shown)


BartoszGrajdek (approved):
LGTM!

PR #419, opened 2024-07-05
fix: check if its first render before updating from props

hannojg:
### Details

See: https://github.com/Expensify/App/issues/43983#issuecomment-2210501859

It's possible that the props aren't updated yet on the JS side, while the native state already is. In that case we'd drop the native state for a stale prop value which is incorrect. So we should make sure to only run this code on the first render.

### Related Issues

https://github.com/Expensify/App/issues/43983
https://github.com/Expensify/App/issues/44185

### Manual Tests

I think its hard to add a specific test for this change.
I verified that the condition is executed on the first render.

### Linked PRs

I don't understand this point 😅

### DIFF
--- a/ios/MarkdownCommitHook.mm
+++ b/ios/MarkdownCommitHook.mm
@@ -141,30 +141,31 @@ RootShadowNode::Unshared MarkdownCommitHook::shadowTreeWillCommit(
                     stateData.attributedStringBox);
 
             // Handles the first render, where the text stored in props is
-            // different than the one stored in state The one in state is empty,
-            // while the one in props is passed from JS If we don't update the
-            // state here, we'll end up with a one-default-line-sized text
-            // input. A better condition to do that can be probably chosen, but
-            // this seems to work
-            auto plainString =
-                std::string([[nsAttributedString string] UTF8String]);
-            if (plainString != textInputProps.text) {
-              // creates new AttributedString from props, adapted from
-              // TextInputShadowNode (ios one, text inputs are
-              // platform-specific)
-              auto attributedString = AttributedString{};
-              attributedString.appendFragment(AttributedString::Fragment{
-                  textInputProps.text, defaultTextAttributes});
+            // different than the one stored in state. The one in state is empty,
+            // while the one in props is passed from JS. If we don't update the
+            // state here, we'll end up with a one-default-line-sized text input
+            if (textInputState.getRevision() == State::initialRevisionValue) {
+                auto plainStringFromState =
+                    std::string([[nsAttributedString string] UTF8String]);
 
-              auto attachments = BaseTextShadowNode::Attachments{};
-              BaseTextShadowNode::buildAttributedString(
-                  defaultTextAttributes, *nodes.textInput, attributedString,
-                  attachments);
+                if (plainStringFromState != textInputProps.text) {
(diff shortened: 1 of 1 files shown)


hannojg:
cc @j-piasecki

ios/MarkdownCommitHook.mm
```diff
@@ -141,14 +141,12 @@
...
             // Handles the first render, where the text stored in props is
-            // different than the one stored in state The one in state is empty,
-            // while the one in props is passed from JS If we don't update the
-            // state here, we'll end up with a one-default-line-sized text
-            // input. A better condition to do that can be probably chosen, but
-            // this seems to work
+            // different than the one stored in state. The one in state is empty,
+            // while the one in props is passed from JS. If we don't update the
+            // state here, we'll end up with a one-default-line-sized text input
             auto plainString =
```
tomekzaw, line 147: There's no need to call `std::string([[nsAttributedString string] UTF8String]);` if `textInputState.getRevision() != State::initialRevisionValue`, can we add another `if` outside to check if it's the first render?

hannojg: true, let me update …

ios/MarkdownCommitHook.mm
```diff
@@ -141,14 +141,12 @@
...
                 std::string([[nsAttributedString string] UTF8String]);
-            if (plainString != textInputProps.text) {
+            if (textInputState.getRevision() == State::initialRevisionValue && plainString != textInputProps.text) {
```
j-piasecki: I think we can even skip comparing state text with props text, but it's been a while since I've touched this code so I'm not sure. We can merge this as-is to fix the issue and investigate it further later.

ios/MarkdownCommitHook.mm
hannojg: like this?

ios/MarkdownCommitHook.mm
hannojg: > I think we can even skip comparing state text with props text,

Yeah I was wondering the same. I think the only case where state == props on first render is when the input is empty 🤔 (but not 100% sure either)

j-piasecki (approved):

ios/MarkdownCommitHook.mm
tomekzaw: yes, exactly, thanks!

tomekzaw:
Unfortunately I get a crash here but it also happens on main :(

Screenshot 2024-07-05 at 11 53 20

hannojg:
let me test on the new arch 🤔 thanks for checking!

tomekzaw:
@hannojg Yep, my fault, see PR https://github.com/Expensify/react-native-live-markdown/pull/420 (we'll merge that in a few minutes)

hannojg:
Cool, do you think when we merge this one soon and could do a new version? asking because I am wondering whether I should create a pr with a patch in newdiot, or If I should just wait a few minutes and then we can do a PR with a version bump?

tomekzaw:
@hannojg Yeah I think we can merge your PR soon, just wanted to test it out. No need for patch right now.

tomekzaw:
@hannojg Could you please merge main?

hannojg:
Done!

PR #426, opened 2024-07-09
Don't use `ShadowNodeFragment::Value`

j-piasecki:
### Details
`ShadowNodeFragment::Value` was removed in https://github.com/facebook/react-native/pull/43962. This PR gets rid of usage of it from the codebase.

### Related Issues

GH_LINK

### Manual Tests
Make sure the library compiles and works on the new architecture.

### DIFF
--- a/cpp/react/renderer/components/RNLiveMarkdownSpec/MarkdownTextInputDecoratorShadowNode.h
+++ b/cpp/react/renderer/components/RNLiveMarkdownSpec/MarkdownTextInputDecoratorShadowNode.h
@@ -11,6 +11,20 @@
 namespace facebook {
 namespace react {
 
+struct OwningShadowNodeFragment {
+  Props::Shared props;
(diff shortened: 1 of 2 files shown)


cpp/react/renderer/components/RNLiveMarkdownSpec/MarkdownTextInputDecoratorShadowNode.h
```diff
@@ -11,6 +11,20 @@
 namespace facebook {
 namespace react {
 
+struct OwningShadowNodeFragment {
+  Props::Shared props;
+  ShadowNode::SharedListOfShared children;
+  State::Shared state;
+  
+  operator ShadowNodeFragment() const {
```
j-piasecki, line 19: Do we want an explicit or implicit conversion here?

tomekzaw (approved):

Issue #317, opened 2024-04-16
Support custom JS parsers without separate bundle


tomekzaw:

tomekzaw:
Proposal: https://swmansion.slack.com/archives/C01GTK53T8Q/p1716235671381059

mallenexpensify:
@tomekzaw do you have the bandwidth to prioritize this? If not, wanna see about tag-teaming it with another engineer or finding a volunteer to take over?  I know you posted the [P/S just yesterday](https://expensify.slack.com/archives/C01GTK53T8Q/p1716235671381059), there was a month+ before that without an update, which is why I'm asking.

tomekzaw:
@mallenexpensify Actually, I have a working proof of concept that uses Reanimated worklets (Reanimated is already a dependency so that's cool) so we might be able to use that quite soon. The only problem for the time being is that ExpensiMark still uses jQuery in some places, but this is something that can be parallelized or even made external, what do you think?

mallenexpensify:
@tomekzaw , honestly, I have no clue. I'm not familiar with this.  Looks like a [plan was hatched in this thread](https://expensify.slack.com/archives/C06BDSWLDPB/p1716410498094849) and that https://github.com/Expensify/App/issues/42494, linked above, is the next step.  thx.

tomekzaw:
Working on it, looks promising, will share a video recording with a demo early next week.

tomekzaw:
Latest update [here](https://swmansion.slack.com/archives/C04878MDF34/p1721225735831989)

szado:
@tomekzaw can you write a short update here for open source users without access to SM Slack?

szt217:
Hello @tomekzaw,

I've been testing out the branch for PR #439 on iOS and the custom parsing is working great. Having an of issue with the Android dev build and I see you were working on it today. Do you have an idea on the timeline for when this PR might be merged?

Appreciate all that you do here.

tomekzaw:
Hi @szt217, thanks for testing out the PR! Happy to hear that custom parsing works well for you.

I've faced some issues with incorrect `MarkdownTextInput` height on RN 0.75 on Android with new arch enabled.

Realistically, I'd expect this PR to be merged next month. There are still some bugs to fix and it needs to be properly reviewed since there's a lot of files changed.

puneetlath:
Hello! Just curious how this is going?

tomekzaw:
[...]

PR #439, opened 2024-07-25
Support custom parsing logic (pass worklet as `parser` prop)

tomekzaw:
### Details
This PR removes separate JS bundle with ExpensiMark logic and allows for customizing `MarkdownTextInput` formatting logic via `parser` property. Parser needs to be a function that accepts a plaintext string and returns an array of ranges.

### Related Issues

https://github.com/Expensify/react-native-live-markdown/issues/317


README.md
```diff
@@ -14,15 +15,17 @@
 First, install the library from npm with the package manager of your choice:
 
 ```sh
-yarn add @expensify/react-native-live-markdown
-npm install @expensify/react-native-live-markdown --save
-npx expo install @expensify/react-native-live-markdown
+yarn add @expensify/react-native-live-markdown react-native-reanimated expensify-common
```
j-piasecki, line 18: Is `expensify-common` needed when using a custom parser? If so, shouldn't it be a peer dependency?

README.md
```diff
@@ -14,11 +15,13 @@
...
+npm install @expensify/react-native-live-markdown react-native-reanimated expensify-common --save
+npx expo install @expensify/react-native-live-markdown react-native-reanimated expensify-common
 ```
 
+React Native Live Markdown requires [react-native-reanimated](https://github.com/software-mansion/react-native-reanimated) 3.11.0 or newer and [expensify-common](https://github.com/Expensify/expensify-common) 2.0.66 or newer.
```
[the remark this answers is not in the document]
tomekzaw: Changed in 0d08556

j-piasecki (approved):
🪨

apple/LiveMarkdownModule.mm
```diff
@@ -1,40 +1,57 @@
...
+  expensify::livemarkdown::injectJSIBindings(rt);
 
-    commitHook_ = std::make_shared<MarkdownCommitHook>(scheduler.uiManager);
-    installed_ = YES;
-  }
-  return @1;
+#ifdef RCT_NEW_ARCH_ENABLED
+  RCTScheduler *scheduler = [surfacePresenter_ scheduler];
+  commitHook_ = std::make_shared<MarkdownCommitHook>(scheduler.uiManager);
+  installed_ = YES;
```
j-piasecki, line 44: I believe this can be removed (and the property as well)

android/src/main/cpp/MarkdownUtils.cpp
```diff
@@ -1,42 +1,30 @@
...
-    jsi::Value result;
-    try {
-      result = func.call(rt, arg);
-    } catch (jsi::JSError e) {
-      result = jsi::Array(rt, 0);
-    }
-    auto json = rt.global().getPropertyAsObject(rt, "JSON").getPropertyAsFunction(rt, "stringify").call(rt, result).asString(rt).utf8(rt);
+      jni::alias_ref<jni::JString> input,
+      int parserId) {
+    const auto markdownRuntime = expensify::livemarkdown::getMarkdownRuntime();
```
j-piasecki, line 14: Is there any chance to somehow indicate here that this method is `synchronized` and mutex isn't necessary here?

apple/LiveMarkdownModule.mm
tomekzaw: Removed in 62bc84e

android/src/main/cpp/MarkdownUtils.cpp
tomekzaw: Comment added in 320e85b

tomekzaw:
Regression: toggling font size doesn't work properly

https://github.com/user-attachments/assets/22ad8085-af79-4e3f-89f7-3da177a26ccc

It works correctly if the text contains a blockquote (`NSParagraphStyle`):

https://github.com/user-attachments/assets/550bf149-5a8f-400a-b00e-b192ec957b20

bbaoNg97:
any update on this?

tomekzaw:
@bbaoNg97 This PR is currently in internal review.

tomekzaw:
TODO: check in release mode

patches/html-entities+2.5.2.patch
```diff
@@ -0,0 +1,8 @@
+diff --git a/node_modules/html-entities/lib/index.js b/node_modules/html-entities/lib/index.js
```
j-piasecki, line 1: I guess this will also be necessary in apps using the lib?

tomekzaw: That's correct, we'll need to patch `html-entities` inside E/App as well.

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -7,78 +7,65 @@
...
-    if (input.equals(mPrevInput)) {
+    if (input.equals(mPrevInput) && mParserId == mPrevParserId) {
       output = mPrevOutput;
     } else {
-      output = parseMarkdown(input);
+      try {
+        output = nativeParseMarkdown(input, mParserId);
+      } catch (Exception e) {
+        output = "[]";
+      }
```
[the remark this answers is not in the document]
tomekzaw: `nativeParseMarkdown` calls `WorkletRuntime::runGuarded` so the error is presented to the user in RedBox. This is just to prevent a native crash. I'd like to improve error handling in a separate PR since this part of the library needs a small refactor and I don't want to spoil this PR with too many changes.

j-piasecki (approved):

garygcchiu:
Hi, thank you for this fix. Unfortunately I think the patch here is causing install issues for yarn users: https://github.com/Expensify/react-native-live-markdown/issues/573

src/parseExpensiMark.ts
```diff
@@ -273,30 +254,24 @@ function groupRanges(ranges: MarkdownRange[]) {
   }, [] as MarkdownRange[]);
 }
 
-function parseExpensiMarkToRanges(markdown: string): MarkdownRange[] {
+function parseExpensiMark(markdown: string): MarkdownRange[] {
   if (markdown.length > MAX_PARSABLE_LENGTH) {
     return [];
   }
-  try {
```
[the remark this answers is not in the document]
tomekzaw: It was a deliberate decision to remove try/catch block so we can show a RedBox with the error if something fails.

However, looks like I forgot about the case when Markdown message before and after parsing don't match. I think we should show a warning and return an empty array in such case, like this:

```tsx
if (text !== markdown) {
  console.warn(
    `[react-native-live-markdown] Parsing error: the processed text does not match the original Markdown input. This may be caused by incorrect parsing functions or invalid input Markdown.\nProcessed input: '${JSON.stringify(
      text,
    )}'\nOriginal input: '${JSON.stringify(markdown)}'`,
  );
  return [];
}
```

PR #464, opened 2024-08-30
Rename type `Range` to `MarkdownRange`

tomekzaw:
### Details
This PR renames type `Range` to `MarkdownRange` to avoid confusion with [type `Range`](https://developer.mozilla.org/en-US/docs/Web/API/Range) from Web API, by the way also used in the codebase.

### Related Issues

GH_LINK


Skalakid (dismissed):
LGTM. Btw in `parserUtils` we have the same `MarkdownRange` type, so it would be nice to merge them into one

BartoszGrajdek (dismissed):

BartoszGrajdek (approved):

PR #469, opened 2024-09-05
Fix `removeSpan` crash on Android

wildan-m:
### Details

When applying formatting, there may be a race condition if we are trying to remove a span that has already been removed.

### Related Issues

https://github.com/Expensify/App/issues/46908

### Manual Tests

1. type: #a
2. double tap a, ensure only a letter selected
3. when tap action shown, choose cut

https://github.com/user-attachments/assets/21bac680-bbd7-43d4-87d3-78baa454a769


wildan-m:
recheck

wildan-m:

tomekzaw:
@wildan-m Thanks for the PR! I will review it soon.

wildan-m:
@tomekzaw Sure, thank you.

Another option could be to include try-catch blocks within `setSpan` and `removeSpan` methods to handle any errors and log them, but it's uncertain if this would be deemed appropriate.

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```java
private void setSpan(SpannableStringBuilder ssb, MarkdownSpan span, int start, int end) {
  try {
    ssb.setSpan(span, start, end, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE);
  } catch (Exception e) {
    Log.e("MarkdownUtils", "Error setting span", e);
  }
}

private void removeSpans(SpannableStringBuilder ssb) {
  // We shouldn't use `removeSpans()` because it also removes SpellcheckSpan, SuggestionSpan etc.
  MarkdownSpan[] spans = ssb.getSpans(0, ssb.length(), MarkdownSpan.class);
  for (MarkdownSpan span : spans) {
    try {
      ssb.removeSpan(span);
    } catch (Exception e) {
      Log.e("MarkdownUtils", "Error removing span: " + span, e);
    }
  }
}
```

tomekzaw:
Thanks again for the PR!

I think we can go even one more step further and get rid of `mShouldSkip` completely:

```tsx
  @Override
  public void beforeTextChanged(CharSequence s, int start, int count, int after) {
    // do nothing
  }

  @Override
  public void onTextChanged(CharSequence s, int start, int before, int count) {
    // do nothing
  }

  @Override
  public void afterTextChanged(Editable editable) {
    if (editable instanceof SpannableStringBuilder) {
      mMarkdownUtils.applyMarkdownFormatting((SpannableStringBuilder) editable);
    }
  }
```

I've tested out this change and it seems to work correctly both on Fabric and Paper.

wildan-m:
@tomekzaw Will we get duplicate calls without the flag?
https://stackoverflow.com/questions/17535415/textwatcher-events-are-being-fired-multiple-times

wildan-m:
@tomekzaw I understand why `applyMarkdownFormatting` is placed in `onTextChanged` to ensure it is only called once for each change.

On each change:
- If we stick to [the current code](https://github.com/Expensify/react-native-live-markdown/blob/2e0bd2c892bb682aebdf00faafb601e165f11112/android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java#L24-L37), `applyMarkdownFormatting` is invoked once.
- In [my approach](https://github.com/Expensify/react-native-live-markdown/blob/15652dca94b376d073ce19cbaf0d099558b5f326/android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java#L24-L44), `applyMarkdownFormatting` is invoked twice.
- Without the `mShouldSkip` flag, `applyMarkdownFormatting` is invoked four times in `afterTextChanged`.

Is it acceptable to use [the second solution](https://github.com/Expensify/react-native-live-markdown/pull/469#issuecomment-2331828762) to ensure `applyMarkdownFormatting` is only called once? beside removeSpan, adding a try-catch block for `setSpan` might be necessary to handle applying formatting to a deleted range. Your thoughts?

wildan-m:
@tomekzaw bump

tomekzaw:
@wildan-m I was curious why `afterTextChanged` is called multiple times and read the Stack Overflow thread you've linked to: https://stackoverflow.com/questions/17535415/textwatcher-events-are-being-fired-multiple-times. My understanding is that whenever the underlying editable changes, `TextWatcher` calls `afterTextChanged`.

One user suggested that setting `android:inputType="textNoSuggestions"` eliminates multiple calls. So I guess that the subsequent calls to `afterTextChanged` are made as a result of text suggestion and spellcheck (which by the way are also treated as spans).

As for the solution with `try/catch` block around `removeSpan` – I guess this just ignores the problem so I'm rather against this approach now.

As for `mShouldSkip` flag – there's an edge case I'd like to notice here – assume that the user has typed "someone@example". This is not a valid email so it will not be underlined. However, the OS can suggest appending ".com" so now the text input value is "someone@example.com" (where `.com` is inside a `SuggestionSpan`) which is a valid e-mail address so this should be underlined. That's why we can't just skip `applyMarkdownFormatting` using `mShouldSkip` flag. 

As for the solution with comparing the previous value (`if (mPreviousEditable != null && mPreviousEditable.toString().equals(editable.toString())`) – note that this just compares the underlying string (without spans). I think this should work correctly in our case so let's proceed with this approach.

android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java
```diff
@@ -22,23 +22,17 @@ public void beforeTextChanged(CharSequence s, int start, int count, int after) {
...
+          if (mPreviousEditable != null && mPreviousEditable.toString().equals(editable.toString())) {
+              return;
+          }
           mMarkdownUtils.applyMarkdownFormatting((SpannableStringBuilder) editable);
-      }
-      
-      // Reset the flag after formatting is applied
-      mShouldSkip = false;
-  }}
+          mPreviousEditable = new SpannableStringBuilder(editable);
```
tomekzaw, line 35: Do we need to store `mPreviousEditable` as a `SpannableStringBuilder`? We just use `mPreviousEditable.toString()` so maybe let's keep it as a string?

wildan-m:
@tomekzaw Thanks for your review. The PR has been updated with unnecessary code removed.

android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java
wildan-m: That's make sense, please wait

wildan-m: done

android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java
```diff
@@ -22,17 +22,18 @@ public void beforeTextChanged(CharSequence s, int start, int count, int after) {
...
   @Override
   public void afterTextChanged(Editable editable) {
-    mShouldSkip = false;
+      if (editable instanceof SpannableStringBuilder) {
+          String currentEditable = editable.toString();
+          if (mPreviousEditable != null && mPreviousEditable.equals(currentEditable)) {
+              return;
+          }
+          mMarkdownUtils.applyMarkdownFormatting((SpannableStringBuilder) editable);
+          mPreviousEditable = currentEditable;
```
tomekzaw, line 36: Let's rename `editable` to `text` to avoid confusion (`editable` contains spans and `text` is just a regular String)
```suggestion
          String currentText = editable.toString();
          if (mPreviousText != null && mPreviousText.equals(currentText)) {
              return;
          }
          mMarkdownUtils.applyMarkdownFormatting((SpannableStringBuilder) editable);
          mPreviousText = currentText;
```

android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java
```diff
@@ -22,17 +22,18 @@ public void beforeTextChanged(CharSequence s, int start, int count, int after) {
...
+      }      
   }
-}
+}
```
tomekzaw, line 39: Let's add a newline at the end of the file

tomekzaw:
@wildan-m I've left some final suggestions

android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java
```diff
@@ -22,17 +22,18 @@ public void beforeTextChanged(CharSequence s, int start, int count, int after) {
...
+          String currentText = editable.toString();
+          if (currentText.equals(mPreviousText)) {
+              return;
+          }
+          mMarkdownUtils.applyMarkdownFormatting((SpannableStringBuilder) editable);
+          mPreviousText = currentText;
+      }      
   }
-}
+}
```
tomekzaw, line 39:
```suggestion
}

```

android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java
```diff
@@ -22,17 +22,18 @@ public void beforeTextChanged(CharSequence s, int start, int count, int after) {
...
+
   }
 
   @Override
   public void afterTextChanged(Editable editable) {
-    mShouldSkip = false;
+    if (editable instanceof SpannableStringBuilder) {
+      String currentText = editable.toString();
+      if (currentText.equals(mPreviousText)) {
+          return;
```
tomekzaw, line 33:
```suggestion
        return;
```

android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java
```diff
@@ -8,8 +8,9 @@
 
 public class MarkdownTextWatcher implements TextWatcher {
   private final MarkdownUtils mMarkdownUtils;
-
+
```
tomekzaw, line 11:
```suggestion

```

tomekzaw (approved):
Thanks for fixing this! I will merge this PR soon.

tomekzaw:
@wildan-m During testing I found one regression – `markdownStyle` changes (e.g. link color) are not reflected until the text is changed:

https://github.com/user-attachments/assets/5a341579-2a55-4b8b-b258-4466b66dc479

This would cause a regression in E/App when toggling light/dark mode.

I was able to fix this by resetting `mPreviousText` in `onTextChanged`:

```diff
   @Override
   public void onTextChanged(CharSequence s, int start, int before, int count) {
+    mPreviousText = null;
   }
```

Here's the correct behavior with this change:

https://github.com/user-attachments/assets/3ec027c7-7ee3-4fb0-8da7-7107ff0b3b62

Alternatively, we could check if `MarkdownStyle` inside `MarkdownUtils` has not changed in the condition.

android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java
```diff
@@ -10,6 +10,7 @@ public class MarkdownTextWatcher implements TextWatcher {
   private final MarkdownUtils mMarkdownUtils;
 
   private boolean mShouldSkip = false;
```
tomekzaw, line 12: Let's also remove this field as it's no longer used.

wildan-m:
> Finally, we can just get rid of this optimization at all

@tomekzaw is it ok to call the apply markdown four times? https://github.com/Expensify/react-native-live-markdown/pull/469#issuecomment-2335002872

> - Without the mShouldSkip flag, applyMarkdownFormatting is invoked four times in afterTextChanged.

checking the regression issue

wildan-m:
@tomekzaw I believe using a `try-catch` solution is the most practical approach here. Ignoring duplicate calls could impact performance, while adding a debounce might introduce noticeable delays even with a 100ms delay.

This issue typically requires a debouncer, but it may not be relevant in this case. Here is the result:

https://github.com/user-attachments/assets/6dfc67de-bea2-4ef9-b5a6-be29c50b9fee

What are your thoughts? is it ok to move with `try-catch` approach?

tomekzaw:
> is it ok to call the apply markdown four times?

@wildan-m I think we can call `applyMarkdownFormatting` several times. If this introduces noticeable performance regressions, we'll address it in a separate issue and separate PR. For now, let's focus on correctness of the formatting. What do you think?

wildan-m:
@tomekzaw resetting `mPreviousText` will call apply formatting twice on each change


https://github.com/user-attachments/assets/eaf3aa09-f245-4181-babe-9bef741d475f


> let's focus on correctness of the formatting. What do you think?

In that case, I suggest keeping the `shouldSkip` flag and reverting to our original solution to minimize unnecessary comparisons. Is that alright with you?

https://github.com/Expensify/react-native-live-markdown/blob/0dfd7b9d6576968f80cdbb59f7144fc5b58afa64/android/src/main/java/com/expensify/livemarkdown/MarkdownTextWatcher.java





This is the debouncer snippet

<details><summary>debounce solution</summary>

```java
package com.expensify.livemarkdown;

import android.os.Handler;
import android.os.Looper;
import android.text.Editable;
import android.text.SpannableStringBuilder;
import android.text.TextWatcher;

import androidx.annotation.NonNull;

public class MarkdownTextWatcher implements TextWatcher {
  private final MarkdownUtils mMarkdownUtils;
  private final Handler mHandler = new Handler(Looper.getMainLooper());
  private final long DEBOUNCE_DELAY = 100; // 300 milliseconds delay
  private Runnable mRunnable;


  public MarkdownTextWatcher(@NonNull MarkdownUtils markdownUtils) {
    mMarkdownUtils = markdownUtils;
  }

  @Override
  public void beforeTextChanged(CharSequence s, int start, int count, int after) {

  }

  @Override
  public void onTextChanged(CharSequence s, int start, int before, int count) {

  }

  @Override
  public void afterTextChanged(final Editable editable) {
    if (editable instanceof SpannableStringBuilder) {
      if (mRunnable != null) {
        mHandler.removeCallbacks(mRunnable);
      }

      mRunnable = new Runnable() {
        @Override
        public void run() {
          System.out.println("[wildebug] applyMarkdownFormatting");
          mMarkdownUtils.applyMarkdownFormatting((SpannableStringBuilder) editable);
        }
      };

      mHandler.postDelayed(mRunnable, DEBOUNCE_DELAY);
    }
  }
}
```

</details>

tomekzaw (changes requested):
Regression found

wildan-m:
@tomekzaw the optimization has been removed

tomekzaw:
@wildan-m Thanks again for your PR. This is now ready to merge. Let's focus on performance improvements in a separate PR.

tomekzaw (approved):
@wildan-m Thanks again for your PR. This is now ready to merge. Let's focus on performance improvements in a separate PR.

tomekzaw:
Regression fixed on Android

https://github.com/user-attachments/assets/ec4320e9-2072-46c2-9e30-7bd5f9d435f2

tomekzaw:
No regressions with height calculation while typing ✅ (content jumps on newlines also happen on `main` branch)

https://github.com/user-attachments/assets/54a934f6-1a28-4b4a-8d80-8da662dda6b9

PR #491, opened 2024-09-16
Fix unsupported type inline image

tomekzaw:
### Details
This PR fixes the following error on Android:

```
Exception in HostFunction: java.lang.IllegalStateException: Unsupported type: inline-image
```

Screenshot 2024-09-16 at 10 30 57

### Related Issues

GH_LINK

### DIFF
--- a/android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
+++ b/android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
@@ -164,8 +164,6 @@ public class MarkdownUtils {
           depth);
         setSpan(ssb, span, start, end);
         break;
-      default:
-        throw new IllegalStateException("Unsupported type: " + type);
     }
   }
 


android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -164,8 +164,6 @@ private void applyRange(SpannableStringBuilder ssb, String type, int start, int
           depth);
         setSpan(ssb, span, start, end);
         break;
-      default:
-        throw new IllegalStateException("Unsupported type: " + type);
```
j-piasecki, line 168: I don't exactly know the use case, but wouldn't replacing this with a warn be less invasive? In case something unexpected ends up here it would still leave a trace. But maybe silently skipping unexpected things really is better, WDYT?

j-piasecki (approved):

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
tomekzaw: Let's just silently ignore things for now (like we do on iOS) and address this idea in a separate PR.

PR #504, opened 2024-09-24
fix: improve performance of RN markdown parsing on large text

dominictb:
### Details

This PR includes minor improvement on `updateTreeElementRefs` function:
  + Remove the block to add the children of every child nodes to the stack. 
  + Query all elements with `data-id` attribute beforehand to improve performance

### Related Issues

https://github.com/Expensify/App/issues/46766

### Manual Tests

1. Paste this text file (over 400k characters) 
[default.txt](https://github.com/user-attachments/files/17113641/default.txt)

2. Observe the performance of the app's input

### Linked PRs

Pending App PR


tomekzaw:
@dominictb Thanks for the PR! The part with max text length looks good to me, but let's wait for @Skalakid's review as for the second part of the PR.

Skalakid (commented):
I think we can split this PR into 2 separate ones. The first one with parser changes and the second one with web implementation improvements.  

Also, we can try adding a `maxLength` prop to web implementation so we can limit how many characters can be entered into the input. For me, it doesn't make sense to allow users to paste text with over 10000 characters. 
The new prop will boost the performance since we will pass only the first `maxLength` letters into our parsing algorithms. What do you think about it?  @dominictb @tomekzaw

tomekzaw:
I don't have an opinion, perhaps we should ask somebody from design team? Also, is there some message length limit on backend?

src/web/utils/treeUtils.ts
```diff
@@ -43,16 +43,22 @@ function addNodeToTree(element: HTMLMarkdownElement, parentTreeNode: TreeNode, t
 
 function updateTreeElementRefs(treeRoot: TreeNode, element: HTMLMarkdownElement) {
   const stack: TreeNode[] = [treeRoot];
+  const treeElements = element.querySelectorAll('[data-id], [data-id] *') as NodeListOf<HTMLMarkdownElement>;
```
Skalakid, line 46:
```suggestion
  const treeElements = element.querySelectorAll('[data-id]') as NodeListOf<HTMLMarkdownElement>;
```

Why we need `[data-id] *`? All elements that can be found in tree have `data-id`

src/web/utils/treeUtils.ts
```diff
@@ -43,16 +43,22 @@ function addNodeToTree(element: HTMLMarkdownElement, parentTreeNode: TreeNode, t
...
+  const treeElementsMapByDataID: Record<string, HTMLMarkdownElement> = {};
+  treeElements.forEach((el) => {
+    if (!el.getAttribute('data-id')) {
+      return;
+    }
+    treeElementsMapByDataID[el.getAttribute('data-id') || ''] = el;
+  });
```
Skalakid, line 53: How about changing `treeElementsMapByDataID` name to `treeElementsMap` or `dataIDToElementMap`?

Skalakid (approved):

tomekzaw (approved):

PR #505, opened 2024-09-26
feat: support maxLength prop in web

dominictb:
### Details

This PR includes the implementation of `maxLength` prop in web

### Related Issues

https://github.com/Expensify/App/issues/46766

### Manual Tests

1. Paste this text file (over 400k characters) 
[default.txt](https://github.com/user-attachments/files/17113641/default.txt)

2. Check that the text input in both web and native is cut off to 30k length

### Linked PRs

Pending App PR


Skalakid:
Hello, I'm testing your changes and I found couple of bugs:

1. The input `value` isn't matching the input content. The `value` (in the picture the text under the input) should be the same as the text inside the component

Screenshot 2024-09-26 at 12 14 49

Skalakid:
2. There are some problems with entering text when we reach the `maxLength` limit. Steps:
 - move the cursor to the end of the text (on the `maxLength` index)
 - start typing
 - notice that the old text is being replaced

https://github.com/user-attachments/assets/3dca11e3-60f3-483e-a366-0cc2fc62fe5f

Skalakid:
Okay, thank you. Now everything works fine :D 

There is one more thing that I wanted to fix there. How about adding value cutting on the first rerender? Currently, when you pass the long text in the `value` prop and the app loads, the `maxLength` prop isn't cutting the input value:

Screenshot 2024-09-26 at 13 22 47

(On the image, I have `maxLength={5}` and default text passed in the `value` prop)

I know the same thing happens on the default RN TextInput, but I think it's worth fixing in our case. What do you think about it?

Skalakid:
Bug: Text isn't being replaced when you reach the limit of `maxLength` prop

https://github.com/user-attachments/assets/a1a7b3fe-05bd-4955-a6ef-b65fe769d520

Expected behavior:


https://github.com/user-attachments/assets/5321189d-80c6-422a-a0a9-5f3330995c94

src/MarkdownTextInput.web.tsx
```diff
@@ -305,14 +316,18 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
...
-        const parsedText = normalizeValue(
+        let parsedText = normalizeValue(
           inputType === 'pasteText' ? pasteContent.current || '' : parseInnerHTMLToText(e.target as MarkdownTextInputElement, inputType, contentSelection.current.start),
         );
 
         if (pasteContent.current) {
           pasteContent.current = null;
         }
 
+        if (typeof maxLength === 'number' && parsedText.length > maxLength) {
```
Skalakid, line 327:
```suggestion
        if (maxLength !== undefined && parsedText.length > maxLength) {
```

src/MarkdownTextInput.web.tsx
```diff
@@ -400,17 +414,25 @@ const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(
         }
 
         const previousText = divRef.current.value;
-        const newText = `${divRef.current.value.substring(0, contentSelection.current.start)}${text}${divRef.current.value.substring(contentSelection.current.end)}`;
+        let insertedText = text;
+        let availableLength = text.length;
+        const prefix = divRef.current.value.substring(0, contentSelection.current.start);
+        const suffix = divRef.current.value.substring(contentSelection.current.end);
+        if (typeof maxLength === 'number') {
```
Skalakid, line 421: Same here

src/MarkdownTextInput.tsx
```diff
@@ -36,15 +36,32 @@ function processMarkdownStyle(input: PartialMarkdownStyle | undefined): Markdown
...
 }
 
-const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>((props, ref) => {
+const MarkdownTextInput = React.forwardRef<TextInput, MarkdownTextInputProps>(({value, maxLength, onChangeText, ...props}, ref) => {
   const IS_FABRIC = 'nativeFabricUIManager' in global;
 
   const markdownStyle = React.useMemo(() => processMarkdownStyle(props.markdownStyle), [props.markdownStyle]);
 
+  const processedText = React.useMemo(() => {
+    if (typeof maxLength === 'number') {
```
Skalakid, line 45:
```suggestion
    if (maxLength === undefined) {
```

src/MarkdownTextInput.tsx
```diff
@@ -36,15 +36,32 @@ function processMarkdownStyle(input: PartialMarkdownStyle | undefined): Markdown
...
+      return value?.slice(0, maxLength);
+    }
+    return value;
+  }, [value, maxLength]);
+
+  React.useLayoutEffect(() => {
+    if (typeof maxLength !== 'number' || !value || value.length <= maxLength || !onChangeText) {
```
Skalakid, line 52:
```suggestion
    if (maxLength === undefined || !value || value.length <= maxLength || !onChangeText) {
```

Skalakid (commented):
Overall, it looks good :D Left some small comments

Skalakid (approved):
LGTM

src/MarkdownTextInput.tsx
```diff
@@ -36,15 +36,35 @@ function processMarkdownStyle(input: PartialMarkdownStyle | undefined): Markdown
...
+    }
+    onChangeTextRef.current(value.slice(0, maxLength));
+  }, [value, maxLength]);
+
   return (
     <>
       <TextInput
         {...props}
+        value={processedText}
+        onChangeText={onChangeText}
```
tomekzaw, line 66: What if the component was supposed to be uncontrolled? I don't really like this logic here

tomekzaw (changes requested):

tomekzaw:
@dominictb Why can't we just pass `maxLength` prop to `TextInput` in the native implementation?

tomekzaw:
@dominictb Thanks for the response

> Basically for value > maxLength, native component will just render the whole value without cutting off anything

Isn't it something we should fix directly in react-native?

tomekzaw:
Actually, a regular `<input type="text">` on web also behaves the same way as currently – if you set `maxlength="10" value="1234567890abcde"`, all 15 characters are visible but you can't edit the text field. I think this is the expected behavior.

Skalakid:
Okay, @tomekzaw convinced me, let's revert changes on the native side and focus on implementing the `maxLength` prop on the web only. Let's make sure it will work the same as the native one and isn't introducing any new bug

tomekzaw:
Basically, `maxLength` prop for `MarkdownTextInput` should work the same way as for `TextInput`.

For Android and iOS, let's just pass this prop to `TextInput` just like we already do (via `{...props}`).

For web, we'll need to implement `maxLength` logic since we're using `contenteditable` which doesn't support `maxlength` out-of-the-box.

All other logic should be moved to E/App.

tomekzaw:
@dominictb Thanks!

Skalakid (dismissed):
LGTM

tomekzaw (approved):

PR #514, opened 2024-10-09
Fix scrolling lines into view

Skalakid:
### Details

THis PR fixies the problem with scrolling elements into view.It changes the logic to now scroll the whole line into view instead of just a child element 
### Related Issues

https://github.com/Expensify/App/issues/50174

### Manual Tests

1. Add a height limit to the input inside the example app to make the component scrollable
2. Clear the input
3. Add newlines so the scroll bar appears on the side of the input
4. Paste emoji 
5. Verify if the emoji is fully visible

### DIFF
--- a/src/web/utils/cursorUtils.ts
+++ b/src/web/utils/cursorUtils.ts
@@ -58,16 +58,12 @@ function scrollIntoView(target: MarkdownTextInputElement, node: TreeNode) {
   const currentLine = target.tree.childNodes[orderIndex]?.element;
   const scrollTargetElement = currentLine || node.element;
 
-  const selection = window.getSelection();
-  if (selection && selection.rangeCount > 0) {
-    const caretRect = scrollTargetElement.getBoundingClientRect();
-    const targetRect = target.getBoundingClientRect();
-
-    // In case the caret is below the visible input area, scroll to the end of the node
-    if (caretRect.top + caretRect.height > targetRect.top + targetRect.height) {
-      targetElement.scrollTop = caretRect.top - targetRect.top + target.scrollTop - targetRect.height + caretRect.height + 4;
-      return;
-    }
+  const caretRect = scrollTargetElement.getBoundingClientRect();
+  const targetRect = target.getBoundingClientRect();
+  // In case the caret is below the visible input area, scroll to the end of the node
+  if (caretRect.top + caretRect.height > targetRect.top + targetRect.height) {
+    targetElement.scrollTop = caretRect.top - targetRect.top + target.scrollTop - targetRect.height + caretRect.height + 4;
+    return;
   }
 
   scrollTargetElement.scrollIntoView({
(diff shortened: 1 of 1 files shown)


src/web/utils/cursorUtils.ts
```diff
@@ -54,28 +54,25 @@ function setCursorPosition(target: MarkdownTextInputElement, startIndex: number,
...
-    }
+  const orderIndex = Number(node.orderIndex.split(',')[0]);
+  const currentLine = target.tree.childNodes[orderIndex]?.element;
+  const scrollTargetElement = currentLine || node.element;
 
-    node.element.scrollIntoView({
-      block: 'nearest',
-    });
+  const selection = window.getSelection();
+  if (selection && selection.rangeCount > 0) {
```
s77rt, line 62: `selection` is no longer being used here. Do we still need this condition?

s77rt (changes requested):

src/web/utils/cursorUtils.ts
Skalakid: Good catch, I forgot to remove it

src/web/utils/cursorUtils.ts
```diff
@@ -54,28 +54,21 @@ function setCursorPosition(target: MarkdownTextInputElement, startIndex: number,
...
-    });
+  const orderIndex = Number(node.orderIndex.split(',')[0]);
+  const currentLine = target.tree.childNodes[orderIndex]?.element;
+  const scrollTargetElement = currentLine || node.element;
+
+  const caretRect = scrollTargetElement.getBoundingClientRect();
+  const targetRect = target.getBoundingClientRect();
+  // In case the caret is below the visible input area, scroll to the end of the node
+  if (caretRect.top + caretRect.height > targetRect.top + targetRect.height) {
+    targetElement.scrollTop = caretRect.top - targetRect.top + target.scrollTop - targetRect.height + caretRect.height + 4;
```
tomekzaw, line 65: Why `+ 4`?

Skalakid: I wanted to leave some space under the line we are scrolling to. It just looks better when we leave some padding. Thanks to that, the text isn't perfectly at the bottom of the input

tomekzaw (approved):

BartoszGrajdek (approved):
lgtm!

src/web/utils/cursorUtils.ts
```diff
@@ -54,28 +54,21 @@ function setCursorPosition(target: MarkdownTextInputElement, startIndex: number,
...
+    return;
```
s77rt, line 66: Do we need this return?

Skalakid: Yes, we only want to change scrollTop. Executing the rest of the code below can cause scroll jumps

Issue #145, opened 2024-01-31
Migrate away from TextKit 1 to TextKit 2


tomekzaw:
The objective of this task is to migrate away from using TextKit 1 APIs in iOS implementation of Live Markdown.

Currently, we access `.layoutManager` property which causes `RCTUITextView` to switch back to TextKit 1 compatibility mode with the following warning:

![warning](https://github.com/Expensify/react-native-live-markdown/assets/20516055/7fe311fa-5f88-4ab1-bee3-462055d19238)

Here's what I did so far:

1. Define class `MarkdownTextLayoutManagerDelegate` that conforms to `NSTextLayoutManagerDelegate`
2. Create an instance of this class in `[MarkdownTextInputDecoratorView didMoveToWindow:]`
3. Store the instance using a strong pointer as a private field
4. Assign an instance of this class to `_textView.textLayoutManager.delegate` (this is a weak pointer!)
5. Define class `MarkdownBlockquoteTextLayoutFragment` that inherits from `NSTextLayoutFragment`
6. Implement `drawAtPoint:inContext:` to draw a rectangle and set a breakpoint inside
7. Implement `textLayoutManager:textLayoutFragmentForLocation:inTextElement:` method that returns an instance of `MarkdownBlockquoteTextLayoutFragment`

This worked successfully, however there was a new issue with alternating content offset on each keypress (see https://github.com/Expensify/react-native-live-markdown/pull/144#issuecomment-1919687702 for details).

Resources:
* https://developer.apple.com/videos/play/wwdc2021/10061/ (tutorial + sample code)

PR #524, opened 2024-10-18
Migrate from `NSLayoutManager` to `NSTextLayoutManager`

tomekzaw:
### Details

Closes https://github.com/Expensify/react-native-live-markdown/issues/145.

### Related Issues

GH_LINK


maksg:
I think that maybe we should consider leaving previous solution as a fallback for iOS versions lower than 16. iOS 16 is a very recent update so it is possible that someone would want to use this on earlier versions.

Other than that this PR looks great in my opinion! 🎉

tomekzaw:
@maksg I've added fallback for iOS 15 and older in 7e56e5e68e158d1fe5c93d221f6a1d28d5b85992

tomekzaw:
| TextKit 1 | TextKit 2 |
|:-:|:-:|
| Screenshot 2025-07-17 at 13 30 22 | Screenshot 2025-07-17 at 13 32 15 |

apple/MarkdownTextLayoutManagerDelegate.mm
```diff
@@ -0,0 +1,22 @@
+#import <RNLiveMarkdown/MarkdownTextLayoutManagerDelegate.h>
+#import <RNLiveMarkdown/BlockquoteTextLayoutFragment.h>
+#import <RNLiveMarkdown/MarkdownFormatter.h>
+
+@implementation MarkdownTextLayoutManagerDelegate
+
+- (NSTextLayoutFragment *)textLayoutManager:(NSTextLayoutManager *)textLayoutManager textLayoutFragmentForLocation:(id <NSTextLocation>)location inTextElement:(NSTextElement *)textElement
+API_AVAILABLE(ios(15.0)){
```
j-piasecki, line 8: Should this be 15 or 16, like in other places?

tomekzaw: Nope, this API is available in iOS 15.0+ according to the docs. This is what Xcode suggested back then when I was writing the code. I know it's weird. https://developer.apple.com/documentation/uikit/nstextlayoutmanagerdelegate/textlayoutmanager(_:textlayoutfragmentfor:in:)?language=objc

PR #534, opened 2024-11-05
Fix emoji formatting

Skalakid:
### Details

This PR adds italics style and strikethrough the emojis on Android and web platforms
### Related Issues

### Manual Tests

1. To the text that contains emojis, add italics styling, ie `_🥰 test 😇 test 🥹_`
2. Verify if the text has italics style and emojis don't
3. Verify if it still works when adding other styles around it, ie `# *~_🥰 test 😇 test 🥹_~*`
4. Verify if emojis aren't strighthrough
5. Test it with different style combinations


android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -66,6 +68,62 @@ public void setMarkdownStyle(@NonNull MarkdownStyle markdownStyle) {
...
+    try {
+      List<MarkdownRange> markdownRanges = new ArrayList<>();
+      JSONArray ranges = new JSONArray(rangesJSON);
+      for (int i = 0; i < ranges.length(); i++) {
+        JSONObject range = ranges.getJSONObject(i);
+        String type = range.getString("type");
+        int start = range.getInt("start");
+        int length = range.getInt("length");
+        int depth = range.optInt("depth", 1);
+        int end = start + length;
```
tomekzaw, line 115: Can we move this calculation to `MarkdownRange` constructor?

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
```diff
@@ -0,0 +1,18 @@
...
+  public int end;
+  public int length;
+  public int depth;
+
+  public MarkdownRange(String type, int start, int end, int depth) {
+    this.type = type;
+    this.start = start;
+    this.end = end;
+    this.length = end - start + 1;
+    this.depth = depth;
```
tomekzaw, line 15:
```suggestion
    this.type = type;
    this.start = start;
    this.length = length;
    this.end = start + length;
    this.depth = depth;
```

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
```diff
@@ -0,0 +1,18 @@
+package com.expensify.livemarkdown;
+
+public class MarkdownRange {
+  public String type;
+  public int start;
+  public int end;
+  public int length;
+  public int depth;
```
tomekzaw, line 8:
```suggestion
  public final String type;
  public final int start;
  public final int end;
  public final int length;
  public final int depth;
```

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -66,6 +68,62 @@ public void setMarkdownStyle(@NonNull MarkdownStyle markdownStyle) {
...
+      }
+    }
+
+    for (int i = 0; i < markdownRanges.size(); i++) {
+      MarkdownRange range = markdownRanges.get(i);
+      if (!range.type.equals("italic") && !range.type.equals("strikethrough")) {
+        continue;
+      }
+
+      for (int j = 0; j <  emojiRanges.size(); j++) {
```
tomekzaw, line 85:
```suggestion
      for (int j = 0; j < emojiRanges.size(); j++) {
```

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -66,6 +68,62 @@ public void setMarkdownStyle(@NonNull MarkdownStyle markdownStyle) {
...
+        MarkdownRange emojiRange = emojiRanges.get(j);
+        if (emojiRange.start >= range.start && emojiRange.end <= range.end) {
+          // Split range
+          MarkdownRange startRange = new MarkdownRange(range.type, range.start, emojiRange.start, range.depth);
+          MarkdownRange endRange =  new MarkdownRange(range.type, emojiRange.end, range.end, range.depth);
```
tomekzaw, line 90:
```suggestion
          MarkdownRange endRange = new MarkdownRange(range.type, emojiRange.end, range.end, range.depth);
```

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -81,22 +139,12 @@ public void applyMarkdownFormatting(SpannableStringBuilder ssb) {
...
-        int start = range.getInt("start");
-        int length = range.getInt("length");
-        int depth = range.optInt("depth", 1);
-        int end = start + length;
-        if (length == 0 || end > input.length()) {
-          continue;
-        }
-        applyRange(ssb, type, start, end, depth);
+    MarkdownRange[] ranges = parseRanges(output);
+    for (MarkdownRange range: ranges) {
```
tomekzaw, line 143:
```suggestion
    for (MarkdownRange range : ranges) {
```

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
tomekzaw, line 85: This algorithm has $O(m*n)$ time complexity where $m$ denotes the number of ranges and $n$ denotes the number of emojis. In worst case scenario, if $m=n$ (i.e. all ranges are emoji ranges), the time complexity is $O(n^2)$. Is there a better algorithm?

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -66,6 +68,62 @@ public void setMarkdownStyle(@NonNull MarkdownStyle markdownStyle) {
...
+      }
+
+      for (int j = 0; j <  emojiRanges.size(); j++) {
+        MarkdownRange emojiRange = emojiRanges.get(j);
+        if (emojiRange.start >= range.start && emojiRange.end <= range.end) {
+          // Split range
+          MarkdownRange startRange = new MarkdownRange(range.type, range.start, emojiRange.start, range.depth);
+          MarkdownRange endRange =  new MarkdownRange(range.type, emojiRange.end, range.end, range.depth);
+          markdownRanges.add(i + 1, endRange);
+          markdownRanges.add(i + 1, startRange);
```
tomekzaw, line 92: `markdownRanges` is an `ArrayList`, wouldn't it be better to use a `LinkedList` if we want to insert elements in the middle of the list?

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -66,6 +68,62 @@ public void setMarkdownStyle(@NonNull MarkdownStyle markdownStyle) {
...
+          markdownRanges.remove(i);
+          i += 1;
```
tomekzaw, line 94: I'm a bit worried about this modification of the loop variable here.

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -66,6 +68,62 @@ public void setMarkdownStyle(@NonNull MarkdownStyle markdownStyle) {
...
+        JSONObject range = ranges.getJSONObject(i);
+        String type = range.getString("type");
+        int start = range.getInt("start");
+        int length = range.getInt("length");
+        int depth = range.optInt("depth", 1);
+        int end = start + length;
+        markdownRanges.add(new MarkdownRange(type, start, end, depth));
+      }
+      List<MarkdownRange> splittedRanges = splitRangesOnEmojis(markdownRanges);
+      return splittedRanges.toArray(new MarkdownRange[0]);
```
tomekzaw, line 119: Why can't we just return `List<MarkdownRange>`?

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -66,6 +68,62 @@ public void setMarkdownStyle(@NonNull MarkdownStyle markdownStyle) {
...
+      for (int i = 0; i < ranges.length(); i++) {
+        JSONObject range = ranges.getJSONObject(i);
+        String type = range.getString("type");
+        int start = range.getInt("start");
+        int length = range.getInt("length");
+        int depth = range.optInt("depth", 1);
+        int end = start + length;
+        markdownRanges.add(new MarkdownRange(type, start, end, depth));
+      }
+      List<MarkdownRange> splittedRanges = splitRangesOnEmojis(markdownRanges);
```
tomekzaw, line 118: There's no need to have these lines of code in `catch (JSONException e)`, the code inside `try` block should be as small as possible.

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
```diff
@@ -81,22 +139,12 @@ public void applyMarkdownFormatting(SpannableStringBuilder ssb) {
...
-        int end = start + length;
-        if (length == 0 || end > input.length()) {
-          continue;
-        }
-        applyRange(ssb, type, start, end, depth);
+    MarkdownRange[] ranges = parseRanges(output);
+    for (MarkdownRange range: ranges) {
+      if (range.length == 0 || range.end > input.length()) {
+        continue;
       }
```
tomekzaw, line 146: Why do we need this? Can't we just make sure we don't add any zero-length ranges?

tomekzaw (changes requested):
Left some suggestions in the comments, also a unit test suite would be much appreciated.

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
Skalakid: Agreed, improved the algorithm time complexity to O(n) in [`8cb08b3` (#534)](https://github.com/Expensify/react-native-live-markdown/pull/534/commits/8cb08b361a81143c1c87aa9c2cf42edacb4afb05)

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
Skalakid: We can, changed

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
Skalakid: Okay

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
Skalakid: Back then, LinkedList might be a better idea, but with the current enhanced O(n) algorithm, we need ArrayList to replace elements in an array of all ranges

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
tomekzaw: This is not resolved

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
tomekzaw: This is not resolved

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
tomekzaw: This is not resolved

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
Skalakid: Forgot to push some changes, now it should be resolved

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
Skalakid: Forgot to push some changes, now it should be resolved

android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
Skalakid: Done in [`6e88d1f` (#534)](https://github.com/Expensify/react-native-live-markdown/pull/534/commits/6e88d1f33aee607ba83ab56fc3de768571cd682c)

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
```diff
@@ -0,0 +1,38 @@
...
+
+  public MarkdownRange(JSONObject range, int index) {
+    try {
+      this.type = range.getString("type");
+      this.start = range.getInt("start");
+      this.length = range.getInt("length");
+      this.end = start + length;
+      this.depth = range.optInt("depth", 1);
+      this.index = index;
+
```
tomekzaw, line 31: Let's remove newlines
```suggestion
```

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
```diff
@@ -0,0 +1,38 @@
...
+    } catch (JSONException e) {
+      throw new RuntimeException(e);
+
```
tomekzaw, line 34:
```suggestion
```

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  public MarkdownRange(JSONObject range, int index) {
+    try {
+      this.type = range.getString("type");
+      this.start = range.getInt("start");
+      this.length = range.getInt("length");
+      this.end = start + length;
+      this.depth = range.optInt("depth", 1);
+      this.index = index;
```
tomekzaw, line 30: `index` shouldn't be a part of `MarkdownRange`.

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
```diff
@@ -0,0 +1,38 @@
...
+  public MarkdownRange(String type, int start, int length, int depth, int index) {
+    this.type = type;
+    this.start = start;
+    this.length = length;
+    this.end = start + length;
+    this.depth = depth;
+    this.index = index;
+  }
+
+  public MarkdownRange(JSONObject range, int index) {
```
tomekzaw, line 23: It's kinda weird that `MarkdownRange` can be constructed only from `JSONObject` and there's no "basic" constructor – but it's a clever approach so I like it. Not sure if constructors are supposed to throw `RuntimeException`, maybe let's move the JSON part back to `parseRanges` and leave only "pure" logic here?

Skalakid: Okay, fixed in https://github.com/Expensify/react-native-live-markdown/pull/534/commits/417ba57a49080096a0fea94a789a1909d2d3a5e4

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
Skalakid: I optimized the current algorithm logic and we don't need `index` anymore. Fixed in https://github.com/Expensify/react-native-live-markdown/pull/534/commits/e963e97366f120ba98f52bba4229b32ae955a139

289Adam289:
I've tested this pr on very example app platform and E/App platforms and I could not find anything. Everything works well including preventing formatting on emojis.

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
```diff
@@ -0,0 +1,21 @@
+package com.expensify.livemarkdown;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
```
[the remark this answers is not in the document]
tomekzaw: Just FYI, I also needed class `MarkdownRange` so I added it in https://github.com/Expensify/react-native-live-markdown/pull/559. I hope you can reuse mine in this PR.

tomekzaw (changes requested):

android/src/main/java/com/expensify/livemarkdown/MarkdownRange.java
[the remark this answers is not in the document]
Skalakid: Done in https://github.com/Expensify/react-native-live-markdown/pull/534/commits/692de7c1ea86765133aa277f3d2635d9ac3c0390

android/src/main/java/com/expensify/livemarkdown/RangeSplitter.java
```diff
@@ -2,11 +2,14 @@
...
 import androidx.annotation.NonNull;
 
+import com.facebook.systrace.Systrace;
+
 import java.util.ArrayList;
 import java.util.List;
 
 public class RangeSplitter {
   public static ArrayList<MarkdownRange> splitRangesOnEmojis(@NonNull List<MarkdownRange> markdownRanges, @NonNull String type) {
+    Systrace.beginSection(0, "splitRangesOnEmojis");
```
tomekzaw, line 12: Please wrap `beginSection`/`endSection` calls with `try/finally` block

android/src/main/java/com/expensify/livemarkdown/RangeSplitter.java
```diff
@@ -41,13 +44,16 @@ public static ArrayList<MarkdownRange> splitRangesOnEmojis(@NonNull List<Markdow
           MarkdownRange newRange = new MarkdownRange(currentRange.getType(), currentStart, emojiStart - currentStart, currentRange.getDepth());
           currentRange = new MarkdownRange(currentRange.getType(), emojiEnd, currentEnd - emojiEnd, currentRange.getDepth());
 
-          newRanges.add(newRange);
+          if (newRange.getLength() > 0) {
+           newRanges.add(newRange);
+          }
```
tomekzaw, line 49: Shouldn't this check be symmetric? So it covers both `_text😀_` and `_😀text_`?

Also, let's rename `newRange` and `currentRange` to `leftRange` and `rightRange`.

Let's avoid creating `newRange` object at all if possible, i.e. move the constructor call into the conditional.

Skalakid:
We decided to close this PR and rewrite the solution to TypeScript at https://github.com/Expensify/react-native-live-markdown/pull/591. After the discussion, we agreed that after enabling custom parsers in the live markdown library, it would be better not to interfere with their results. Because of that, we moved the range splitting algorithm to the `parseExpensiMark` so all platforms get the split ranges. It also gives a choice in custom parser implementation if someone wants to remove styles from emojis

PR #537, opened 2024-11-05
Improve removing spans on Android

tomekzaw:
### Details

This PR changes the way how spans are removed before applying new formatting.

Previously, we would call `ssb.getSpans(0, ssb.length(), MarkdownSpan.class);` to find all span added in the previous run. All added spans implement `MarkdownSpan` interface.

After this change, each time we add a span, we add it to `mMarkdownSpans` list (`ArrayList`). Then, we iterate over that list to remove all previously added spans and clear the list afterwards.

TODO: test performance

### Related Issues

GH_LINK

### DIFF
--- a/android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
+++ b/android/src/main/java/com/expensify/livemarkdown/MarkdownUtils.java
@@ -16,6 +16,7 @@ import org.json.JSONObject;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.Objects;
 
 public class MarkdownUtils {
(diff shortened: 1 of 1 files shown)


tomekzaw:
TODO: test performance

j-piasecki (dismissed):
The code looks good but what's the reasoning behind this change?

tomekzaw:
There's no particular reason for this change right now. Android spans are known to be slow and I want to see if we can make it faster by eliminating `ssb.getSpan` call in favor of keeping a list of spans.

tomekzaw:
I've tested out this change on Google Pixel 4 in release mode with profiler attached and the performance is comparable (`removeSpans` takes 1.09 ms vs. 1.12 ms). In debug mode, `removeSpans` takes ~3ms.

PR #541, opened 2024-11-07
@jpiasecki/memoize utils commit hook

j-piasecki:
### Details
Instead of creating `RCTMarkdownUtils` every time the commit hook is run, memoize it in the state of the decorator shadow node. We can accomplish that by cloning the decorator node after the text input node is cloned. This way, we can clone the decorator in-place since the parent would be mutable at that point in time.

### Manual Tests
Verify that the library still works and isn't flickering on the new architecture


tomekzaw (approved):
Thanks for this PR!

PR #547, opened 2024-11-18
Fix text selection on Web doesn't work when passing markdownStyle without useMemo

SzymczakJ:
### Related Issues

GH_LINK

### DIFF
--- a/src/styleUtils.ts
+++ b/src/styleUtils.ts
@@ -89,10 +89,43 @@ function mergeMarkdownStyleWithDefault(input: PartialMarkdownStyle | undefined):
   return output;
 }
 
+function deepEqualMarkdownStyles(markdownStyle1: MarkdownStyle, markdownStyle2: PartialMarkdownStyle) {
+  const keys1 = Object.keys(markdownStyle1) as Array<keyof MarkdownStyle>;
+  const keys2 = Object.keys(markdownStyle2) as Array<keyof MarkdownStyle>;
+
+  if (keys1.length !== keys2.length) {
+    return false;
+  }
+
+  let areStylesEqual = true;
+  keys1.forEach((key) => {
+    // Because each of the markdown objects types are not overlapping, we have to case to Record<string, string | number>
+    const singleStyle1 = markdownStyle1[key] as Record<string, string | number>;
+    const singleStyle2 = markdownStyle2[key] as Record<string, string | number>;
+    const singleStyleKeys1 = Object.keys(singleStyle1);
+    const singleStyleKeys2 = Object.keys(singleStyle2);
+    if (singleStyleKeys1.length !== singleStyleKeys2.length) {
+      areStylesEqual = false;
+      return;
+    }
+    singleStyleKeys1.forEach((styleKey) => {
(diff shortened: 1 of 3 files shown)


src/styleUtils.ts
```diff
@@ -89,10 +89,43 @@ function mergeMarkdownStyleWithDefault(input: PartialMarkdownStyle | undefined):
   return output;
 }
 
+function deepEqualMarkdownStyles(markdownStyle1: MarkdownStyle, markdownStyle2: PartialMarkdownStyle) {
+  const keys1 = Object.keys(markdownStyle1) as Array<keyof MarkdownStyle>;
+  const keys2 = Object.keys(markdownStyle2) as Array<keyof MarkdownStyle>;
+
+  if (keys1.length !== keys2.length) {
+    return false;
+  }
```
tomekzaw, line 98: Since all values of `markdownStyle` are serializable, why can't we just compare `JSON.stringify(markdownStyle1) === JSON.stringify(markdownStyle2)`?

src/styleUtils.ts
```diff
@@ -89,10 +89,43 @@ function mergeMarkdownStyleWithDefault(input: PartialMarkdownStyle | undefined):
...
+}
+
 function parseStringWithUnitToNumber(value: string | null): number {
   return value ? parseInt(value.replace('px', ''), 10) : 0;
 }
 
 export type {PartialMarkdownStyle};
 
-export {mergeMarkdownStyleWithDefault, parseStringWithUnitToNumber};
+export {mergeMarkdownStyleWithDefault, parseStringWithUnitToNumber, deepEqualMarkdownStyles as deepCompareMarkdownStyles};
```
tomekzaw, line 131: Why do we export `deepEqualMarkdownStyles` with a different name `deepCompareMarkdownStyles`?

tomekzaw (changes requested):
Also, does the selection work correctly when changing `markdownStyle`?

src/styleUtils.ts
SzymczakJ: Don't we want to treat objects like `{color: '#FF0000', backgroundColor: '#FF0000'}` and `{backgroundColor: '#FF0000', color: '#FF0000'}` as equal? That's the only reason I didn't use `JSON.stringify`, because it maintains order. We could use `JSON.stringify`, but then some logically equal objects(but with changed order of props) would come out as not equal. WDYT?

src/styleUtils.ts
SzymczakJ: Don't we want to treat objects like `{color: '#FF0000', backgroundColor: '#FF0000'}` and `{backgroundColor: '#FF0000', color: '#FF0000'} `as equal? That's the only reason I didn't use `JSON.stringify`, because it maintains order. We could use `JSON.stringify`, but then some logically equal objects(but with changed order of props) would come out as not equal. WDYT?

BartoszGrajdek:
Maybe we should write some tests that would cover this case? @tomekzaw

jmusial:
Can be closed, has been fixed in https://github.com/Expensify/react-native-live-markdown/pull/684

PR #520, opened 2024-10-17
Use `NSTextStorageDelegate` instead of method swizzling

tomekzaw:
### Details

Work in progress.

This PR is supposed to fix these issues:
* https://github.com/Expensify/react-native-live-markdown/issues/34
* https://github.com/Expensify/react-native-live-markdown/issues/101
* https://github.com/Expensify/react-native-live-markdown/issues/181
* https://github.com/Expensify/react-native-live-markdown/issues/309
* https://github.com/Expensify/react-native-live-markdown/issues/417
* https://github.com/Expensify/react-native-live-markdown/issues/452
* https://github.com/Expensify/react-native-live-markdown/issues/540
* https://github.com/Expensify/react-native-live-markdown/issues/646

### Related Issues

GH_LINK

PR #556, opened 2024-12-04
Handle context menu format bold and italic

bernhardoj:
### Details

Handle the context menu Format command for Bold and Italic.

### Related Issues

https://github.com/Expensify/App/issues/53145

### Manual Tests

1. Run the web example app on Safari web/mWeb
2. Type anything on the text input
3. Select any of the text
4. Right-click/long-press and choose Format context menu option
5. Choose Bold or Italic
6. Verify it wraps the selected text with the selected markdown (* for Bold, _ for Italic)

https://github.com/user-attachments/assets/2df8cf2f-a784-482e-be99-9f38d52cb086

https://github.com/user-attachments/assets/9959a8a3-9cf7-4561-aa9d-740f23b02e96

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -15,6 +15,18 @@ import {PlatformInfo} from './PlatformInfo';
 // We don't need this workaround in New Expensify App since Reanimated is imported before Live Markdown.
 console.log(Animated);
 
+function handleFormatSelection(selectedText: string, formatCommand: string) {
+  if (formatCommand === 'formatBold') {
+    return `*${selectedText}*`;
+  }
+
+  if (formatCommand === 'formatItalic') {
+    return `_${selectedText}_`;
+  }
+
+  return selectedText;
+}
+
 export default function App() {
   const [value, setValue] = React.useState(TEST_CONST.EXAMPLE_CONTENT);
   const [textColorState, setTextColorState] = React.useState(false);
@@ -48,7 +60,7 @@ export default function App() {
       <PlatformInfo />
       <MarkdownTextInput
         multiline
-        formatSelection={(selectedText, formatCommand) => formatCommand === 'formatBold' ? `*${selectedText}*` : formatCommand === 'formatItalic' ? `_${selectedText}_` : selectedText}
+        formatSelection={handleFormatSelection}
         autoCapitalize="none"
         value={value}
         onChangeText={setValue}
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -16,15 +16,14 @@ import {PlatformInfo} from './PlatformInfo';
 console.log(Animated);
 
 function handleFormatSelection(selectedText: string, formatCommand: string) {
-  if (formatCommand === 'formatBold') {
-    return `*${selectedText}*`;
+  switch (formatCommand) {
+    case 'formatBold':
+      return `*${selectedText}*`;
+    case 'formatItalic':
+      return `_${selectedText}_`;
+    default:
+      return selectedText;
   }
-
-  if (formatCommand === 'formatItalic') {
-    return `_${selectedText}_`;
-  }
-
-  return selectedText;
 }
 
 export default function App() {
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -48,6 +48,7 @@ export default function App() {
       <PlatformInfo />
       <MarkdownTextInput
         multiline
+        formatSelection={(selectedText, formatCommand) => formatCommand === 'formatBold' ? `*${selectedText}*` : formatCommand === 'formatItalic' ? `_${selectedText}_` : selectedText}
         autoCapitalize="none"
         value={value}
         onChangeText={setValue}
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -240,7 +240,7 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
 
     const handleFormatSelection = useCallback(
       (target: MarkdownTextInputElement, parsedText: string, cursorPosition: number, formatCommand: string): ParseTextResult => {
-        if (!contentSelection.current) {
+        if (!contentSelection.current || contentSelection.current.end - contentSelection.current.start < 1) {
           return {
             text: '',
             cursorPosition: 0,
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -260,14 +260,19 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
             break;
         }
 
-        const beforeSelectedText = parsedText.slice(0, contentSelection.current.start);
         const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
         const formattedText = formatSelection?.(selectedText, formatType) ?? selectedText;
-        const formattedTextDiffLength = formattedText.length - selectedText.length;
-        const afterSelectedText = parsedText.slice(contentSelection.current.end);
-        const text = `${beforeSelectedText}${formattedText}${afterSelectedText}`;
 
-        return parseText(parser, target, text, processedMarkdownStyle, cursorPosition + formattedTextDiffLength, true);
+        if (selectedText === formattedText) {
+          return parseText(parser, target, parsedText, processedMarkdownStyle, cursorPosition);
+        }
+
+        const prefix = parsedText.slice(0, contentSelection.current.start);
+        const suffix = parsedText.slice(contentSelection.current.end);
+        const diffLength = formattedText.length - selectedText.length;
+        const text = `${prefix}${formattedText}${suffix}`;
+
+        return parseText(parser, target, text, processedMarkdownStyle, cursorPosition + diffLength, true);
       },
       [parser, parseText, formatSelection, processedMarkdownStyle],
     );
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -241,10 +241,7 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
     const handleFormatSelection = useCallback(
       (target: MarkdownTextInputElement, parsedText: string, cursorPosition: number, formatCommand: string): ParseTextResult => {
         if (!contentSelection.current || contentSelection.current.end - contentSelection.current.start < 1) {
-          return {
-            text: '',
-            cursorPosition: 0,
-          };
+          throw new Error('[react-native-live-markdown] invalid selection');
         }
 
         const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -244,8 +244,12 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
           throw new Error('[react-native-live-markdown] Trying to apply format command on empty selection');
         }
 
+        if (!formatSelection) {
+          return parseText(parser, target, parsedText, processedMarkdownStyle, cursorPosition);
+        }
+
         const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
-        const formattedText = formatSelection?.(selectedText, formatCommand) ?? selectedText;
+        const formattedText = formatSelection(selectedText, formatCommand);
 
         if (selectedText === formattedText) {
           return parseText(parser, target, parsedText, processedMarkdownStyle, cursorPosition);
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -241,7 +241,7 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
     const handleFormatSelection = useCallback(
       (target: MarkdownTextInputElement, parsedText: string, cursorPosition: number, formatCommand: string): ParseTextResult => {
         if (!contentSelection.current || contentSelection.current.end - contentSelection.current.start < 1) {
-          throw new Error('[react-native-live-markdown] invalid selection');
+          throw new Error('[react-native-live-markdown] Trying to apply format command on empty selection');
         }
 
         const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -18,20 +18,19 @@ import {updateInputStructure} from './web/utils/parserUtils';
 import InputHistory from './web/InputHistory';
 import type {TreeNode} from './web/utils/treeUtils';
 import {getCurrentCursorPosition, removeSelection, setCursorPosition} from './web/utils/cursorUtils';
-import {getFormatType} from './web/utils/blockUtils';
 import './web/MarkdownTextInput.css';
 import type {MarkdownStyle} from './MarkdownTextInputDecoratorViewNativeComponent';
 import {getElementHeight, getPlaceholderValue, isEventComposing, normalizeValue, parseInnerHTMLToText} from './web/utils/inputUtils';
 import {parseToReactDOMStyle, processMarkdownStyle} from './web/utils/webStyleUtils';
 import {forceRefreshAllImages} from './web/inputElements/inlineImage';
-import type {MarkdownRange, InlineImagesInputProps, FormatType} from './commonTypes';
+import type {MarkdownRange, InlineImagesInputProps} from './commonTypes';
 
 const useClientEffect = typeof window === 'undefined' ? useEffect : useLayoutEffect;
 
 interface MarkdownTextInputProps extends TextInputProps, InlineImagesInputProps {
   markdownStyle?: MarkdownStyle;
   parser: (text: string) => MarkdownRange[];
-  formatSelection?: (selectedText: string, formatType: FormatType) => string;
+  formatSelection?: (selectedText: string, formatCommand: string) => string;
   onClick?: (e: MouseEvent<HTMLDivElement>) => void;
   dir?: string;
   disabled?: boolean;
@@ -248,9 +247,8 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
           };
         }
 
-        const formatType = getFormatType(formatCommand);
         const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
-        const formattedText = formatSelection?.(selectedText, formatType) ?? selectedText;
+        const formattedText = formatSelection?.(selectedText, formatCommand) ?? selectedText;
 
         if (selectedText === formattedText) {
           return parseText(parser, target, parsedText, processedMarkdownStyle, cursorPosition);
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -238,7 +238,7 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
       [parser, parseText, processedMarkdownStyle],
     );
 
-    const format = useCallback(
+    const handleFormatSelection = useCallback(
       (target: MarkdownTextInputElement, parsedText: string, cursorPosition: number, formatCommand: string): ParseTextResult => {
         if (!contentSelection.current) {
           return {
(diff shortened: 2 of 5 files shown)


tomekzaw:
Hi @bernhardoj, thanks for submitting this PR. I'm okay with the changes.

However, this PR only implements Format menu on web. I believe the same feature is missing on native iOS. Are there any plans to implement this missing feature also in native iOS?

bernhardoj:
I don't see the Format option on iOS native. I think it's only available on the web.

https://github.com/user-attachments/assets/df3f9386-3f69-4685-a066-241c3c8345cf

tomekzaw:
@bernhardoj You're right, if there's no menu on native iOS then we don't need to add one right now.

In that case I think we can proceed with web-only implementation.

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +236,35 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
+        let markdown;
+        switch (formatType) {
+          case 'formatBold':
+            markdown = '*';
+            break;
+          case 'formatItalic':
+            markdown = '_';
+            break;
+          default:
+            markdown = '';
```
tomekzaw, line 256: If `formatType` is other than `formatBold` or `formatItalic`, then we'll still call the code below. Is it possible to early-return here instead of running the code below with `markdown = ''`?

tomekzaw, line 256: I think this state is unreachable so we could just throw an error here.

Skalakid:
Hello @bernhardoj, thank you for your PR! 

Yesterday, we merged support for adding custom parsers to the Live Markdown Input. It's a pretty big change, and it extends the use cases for our library. Because of that, I think we should slightly update your solution in this PR. Now you've implemented adding static syntaxes in the `format` function, like: `*` for bold and `_` for italics. However, because now users can use other custom parsers together with the Live Markdown library, it would be perfect to allow them to choose other custom syntaxes for their styles. They can implement a parser that `/sample text/` string into bold and use `/` instead of `*`.``
To solve this problem, how about adding the `formatSelection` prop to the Live Markdown Input, so users can handle this case on their own? In the case of Expensify, it will look something like this:

``` js
formatSelection={(selectedText: string, style: 'bold' | 'italic' | 'underline') => style === 'bold' ? `*${selectedText}*` : style === 'italic' ? `_${selectedText}_` : selectedText}
```

What do you think about it?

bernhardoj:
That sounds good, I'll update it tomorrow!

src/MarkdownTextInput.web.tsx
bernhardoj: Is it better if I add underline as the default value?
```
switch (formatCommand) {
  case 'formatBold':
    formatType = 'bold';
    break;
  case 'formatItalic':
    formatType = 'italic';
    break;
  default:
    formatType = 'underline';
    break;
}
```

bernhardoj:
@Skalakid I need help with defining the type for `formatSelection`. Since this is only for web, is it fine to add the type to `MarkdownTextInputProps` on both `MarkdownTextInput.tsx` and `MarkdownTextInput.web.tsx` files?
https://github.com/Expensify/react-native-live-markdown/blob/9ad0acdc7ffcd2d0be23fb64031bcac2ebed872e/src/MarkdownTextInput.tsx#L54-L57
https://github.com/Expensify/react-native-live-markdown/blob/9ad0acdc7ffcd2d0be23fb64031bcac2ebed872e/src/MarkdownTextInput.web.tsx#L30-L36

```
formatSelection?: (selectedText: string, formatType: FormatType) => string;
```

tomekzaw:
@bernhardoj Just FYI, I was able to add "Format" button in context menu on native iOS with the following code snippet:

```objc
// inside UITextView
UIMenuItem *formatItem = [[UIMenuItem alloc] initWithTitle:@"Format" action:@selector(command:textView:)];
UIMenuController.sharedMenuController.menuItems = @[formatItem];
```

```objc
- (void)command:(nonnull UICommand *)command textView:(nonnull UITextView *)textView
{
  // Do something
}
```

bernhardoj:
Should I add it here?
https://github.com/Expensify/react-native-live-markdown/blob/5695408fe6286724276438e4e957ed2a6b2101c3/apple/RCTUITextView%2BMarkdown.mm#L29-L40

Also, I get this error when trying to run the iOS example app.
> error export CLANG_WARN_DOCUMENTATION_COMMENTS\=YES
error export CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER\=NO
error export GCC_WARN_UNDECLARED_SELECTOR\=YES
error export VALIDATE_PRODUCT\=NO
error Failed to build ios project. "xcodebuild" exited with error code '65'. To debug build logs further, consider building your app with Xcode.app, by opening 'LiveMarkdownExample.xcworkspace'.
npm ERR! Lifecycle script `ios` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: @expensify/react-native-live-markdown-example@0.0.1 
npm ERR!   at location: /Users/bernhardoj/Documents/dev/freelance/react-native-live-markdown/example

tomekzaw:
@bernhardoj I think we can implement native iOS in a separate PR. Once we merge https://github.com/Expensify/react-native-live-markdown/pull/520 it should be a bit simpler to implement that.

tomekzaw:
@bernhardoj Also, I'm curious what's the expected behavior when `*foo*` is selected and "Bold" button is pressed. I would assume that formatting should be removed then, i.e. the asterisks should be removed.

bernhardoj:
> I would assume that formatting should be removed then, i.e. the asterisks should be removed.

I tested using the `contenteditable` and it indeed removes the bold when we select the bold format again. Also, Safari nicely shows that the Bold format is selected (I noticed that it shows the Bold as selected as long as the selected text has a bold font-weight style).
image

I wonder how we are going to handle the "unwrap" since we now allow a custom parser and handle cases like only selecting part of the bolded text (`*b|ol|d*`) or a combination of bolded and non-bolded texts (`*bo|ld* nor|mal`).

Also, the header markdown shows Bold format as selected.
image

Skalakid:
> Since this is only for web, is it fine to add the type to MarkdownTextInputProps on both MarkdownTextInput.tsx and MarkdownTextInput.web.tsx files?

@bernhardoj I think, yes. You can add it to both files. Thank to that there won't be any TS errors in the app

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,40 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
       [parser, parseText, processedMarkdownStyle],
     );
 
+    const format = useCallback(
```
Skalakid, line 241: How about changing this function name to something like `handleFormatSelection` or `handleFormatSelectedText`? I think the `format` name is too general.

Skalakid (changes requested):

src/MarkdownTextInput.web.tsx
bernhardoj: That makes sense. Updated

bernhardoj:
@Skalakid hi, how's the review going?

Skalakid:
Hi, I reviewed the PR, and the code looks good to me, however, together with @tomekzaw, we noticed that this issue is more complex than we thought. The current implementation allows to modify the selected text, however, there are many other cases connected to removing style from the text, that should be handled. Here is a list of them:

- Removing style when all the text between style syntaxes is selected: 
`*[world]*` -> `[world]`
In this case the `"*"` syntaxes from outside the text selection, should be removed. In current implementation we are only returning the currently selected text so it will be hard to remove them

- Removing style from the middle of the text:
`*hello [world] test*` -> `*hello* [world] *test*` 
The same situation here, with by only having the selected text (`world`) it would be hard to split the styled text in the middle

I think we should reconsider the current solution and create simple to use API for `react-native-live-markdown` library that will allow to implement style formatting in both directions (adding and removing the style). How about adding more props to the `formatSelection` function like: selection start, selection end, active markdown ranges in the current selection

Skalakid:
On the other hand, I checked now, and WhatsApp doesn't support removing the styles in the format option.


https://github.com/user-attachments/assets/fb568b76-61d2-4cf1-93a6-af3b8e0ad31c

So I think we can merge it as it is and enhance the `formatSelection ` function in the future

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,40 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
+          case 'formatBold':
+            formatType = 'bold';
+            break;
+          case 'formatItalic':
+            formatType = 'italic';
+            break;
+          default:
+            formatType = 'underline';
+            break;
+        }
```
Skalakid, line 261: Why `underline` is set as default?

Skalakid, line 261: Let's move this switch elsewhere, for example to `blockUtils.ts`

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,40 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
+
+        const beforeSelectedText = parsedText.slice(0, contentSelection.current.start);
```
Skalakid, line 263:
```suggestion
        const prefix = parsedText.slice(0, contentSelection.current.start);
```

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,40 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
+        const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
+        const formattedText = formatSelection?.(selectedText, formatType) ?? selectedText;
+        const formattedTextDiffLength = formattedText.length - selectedText.length;
+        const afterSelectedText = parsedText.slice(contentSelection.current.end);
```
Skalakid, line 267:
```suggestion
        const suffix = parsedText.slice(contentSelection.current.end);
```

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,40 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
+            break;
+          default:
+            formatType = 'underline';
+            break;
+        }
+
+        const beforeSelectedText = parsedText.slice(0, contentSelection.current.start);
+        const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
+        const formattedText = formatSelection?.(selectedText, formatType) ?? selectedText;
+        const formattedTextDiffLength = formattedText.length - selectedText.length;
```
Skalakid, line 266:
```suggestion
        const diffLength = formattedText.length - selectedText.length;
```

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,40 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
+            formatType = 'italic';
+            break;
+          default:
+            formatType = 'underline';
+            break;
+        }
+
+        const beforeSelectedText = parsedText.slice(0, contentSelection.current.start);
+        const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
+        const formattedText = formatSelection?.(selectedText, formatType) ?? selectedText;
```
Skalakid, line 265: We can check if the returned text actually changed after running `formatSelection`, to avoid unnecessary text parsing

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,40 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
     );
 
+    const handleFormatSelection = useCallback(
+      (target: MarkdownTextInputElement, parsedText: string, cursorPosition: number, formatCommand: string): ParseTextResult => {
+        if (!contentSelection.current) {
+          return {
+            text: '',
+            cursorPosition: 0,
+          };
+        }
```
Skalakid, line 248: How about early returning also when contentSelection length between start and end is less than 1?

Skalakid (changes requested):

src/MarkdownTextInput.web.tsx
bernhardoj: Done

src/MarkdownTextInput.web.tsx
bernhardoj: Done

src/MarkdownTextInput.web.tsx
bernhardoj: Done

src/MarkdownTextInput.web.tsx
bernhardoj: Done

src/MarkdownTextInput.web.tsx
bernhardoj: Done

src/MarkdownTextInput.web.tsx
bernhardoj: Done

src/MarkdownTextInput.web.tsx
bernhardoj: I actually asked about this [here](https://github.com/Expensify/react-native-live-markdown/pull/556#discussion_r1873492458). Since we only handle the bold, italic, and underline command, I default it to underline. Should I use the "throw" approach?

blimpich:
@Skalakid @tomekzaw friendly bump on re-reviewing this PR 🙂

mananjadhav:
@Skalakid @tomekzaw quick bump on this one.

Skalakid (commented):
Sorry for the delay, we were OOO.
@bernhardoj The current code looks good to me, but can you add `formatSelection` to the example app, please?

Also, [here](https://github.com/Expensify/react-native-live-markdown/pull/556#discussion_r1905063706) I left one more suggestion ;)

src/web/utils/blockUtils.ts
```diff
@@ -110,4 +110,15 @@ function extendBlockStructure(
...
 
-export {addStyleToBlock, extendBlockStructure, isBlockMarkdownType, getFirstBlockMarkdownRange};
+function getFormatType(formatCommand: string): FormatType {
+  switch (formatCommand) {
+    case 'formatBold':
+      return 'bold';
+    case 'formatItalic':
+      return 'italic';
+    default:
+      return 'underline';
```
tomekzaw, line 120: Not sure why underline is the default. This is not future-proof as there could be other styles as well. Maybe let's just remove this additional layer of conversion and pass `formatCommand` directly to the `formatSelection` function? This way it will be more generic.

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +239,33 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
       [parser, parseText, processedMarkdownStyle],
     );
 
+    const handleFormatSelection = useCallback(
+      (target: MarkdownTextInputElement, parsedText: string, cursorPosition: number, formatCommand: string): ParseTextResult => {
+        if (!contentSelection.current || contentSelection.current.end - contentSelection.current.start < 1) {
+          return {
+            text: '',
+            cursorPosition: 0,
+          };
```
tomekzaw, line 248: When exactly will this case happen? If this case is not expected, wouldn't it be better to throw an error instead?

src/MarkdownTextInput.web.tsx
```diff
@@ -18,18 +18,20 @@ import {updateInputStructure} from './web/utils/parserUtils';
...
 import {forceRefreshAllImages} from './web/inputElements/inlineImage';
-import type {MarkdownRange, InlineImagesInputProps} from './commonTypes';
+import type {MarkdownRange, InlineImagesInputProps, FormatType} from './commonTypes';
 
 const useClientEffect = typeof window === 'undefined' ? useEffect : useLayoutEffect;
 
 interface MarkdownTextInputProps extends TextInputProps, InlineImagesInputProps {
   markdownStyle?: MarkdownStyle;
   parser: (text: string) => MarkdownRange[];
+  formatSelection?: (selectedText: string, formatType: FormatType) => string;
```
tomekzaw, line 34: Can't we just pass the name of the format command, i.e. `"formatBold"` or `"formatItalic"` to the function instead converting to `"bold"` or `"italic"`?

src/MarkdownTextInput.web.tsx
bernhardoj: I don't think it's possible to happen. Updated to throw an error.

src/web/utils/blockUtils.ts
bernhardoj: Updated

src/MarkdownTextInput.web.tsx
bernhardoj: Updated

src/MarkdownTextInput.web.tsx
bernhardoj: Since we don't do the conversion anymore, no throw is needed here.

example/src/App.tsx
```diff
@@ -48,6 +48,7 @@ export default function App() {
       <PlatformInfo />
       <MarkdownTextInput
         multiline
+        formatSelection={(selectedText, formatCommand) => formatCommand === 'formatBold' ? `*${selectedText}*` : formatCommand === 'formatItalic' ? `_${selectedText}_` : selectedText}
```
bernhardoj, line 51: > but can you add formatSelection to the example app, please?

Already added it here.

tomekzaw: Can we wrap this function with `useCallback` or even move it to top-level scope (as `function formatSelection(...) { ... }`)?

example/src/App.tsx
```diff
@@ -15,6 +15,18 @@ import {PlatformInfo} from './PlatformInfo';
...
+function handleFormatSelection(selectedText: string, formatCommand: string) {
+  if (formatCommand === 'formatBold') {
+    return `*${selectedText}*`;
+  }
+
+  if (formatCommand === 'formatItalic') {
+    return `_${selectedText}_`;
+  }
+
+  return selectedText;
```
tomekzaw, line 27: This is a nit but this is a perfect place to use `switch/case` block.
```suggestion
  switch (formatCommand) {
    case 'formatBold':
      return `*${selectedText}*`;
    case 'formatItalic':
      return `_${selectedText}_`;
    default:
      return selectedText;
  }
```

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,29 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
       [parser, parseText, processedMarkdownStyle],
     );
 
+    const handleFormatSelection = useCallback(
+      (target: MarkdownTextInputElement, parsedText: string, cursorPosition: number, formatCommand: string): ParseTextResult => {
+        if (!contentSelection.current || contentSelection.current.end - contentSelection.current.start < 1) {
+          throw new Error('[react-native-live-markdown] invalid selection');
```
tomekzaw, line 244: Let's rephrase the error message to be more specific.
```suggestion
          throw new Error('[react-native-live-markdown] Trying to apply format command on empty selection');
```

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,29 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
+        }
+
+        const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
+        const formattedText = formatSelection?.(selectedText, formatCommand) ?? selectedText;
```
tomekzaw, line 248: Why do we need the `?? selectedText` part here? Maybe we should just return early if `formatSelection` is undefined?

src/MarkdownTextInput.web.tsx
bernhardoj: Done

src/MarkdownTextInput.web.tsx
```diff
@@ -236,6 +238,29 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
+
+        if (selectedText === formattedText) {
+          return parseText(parser, target, parsedText, processedMarkdownStyle, cursorPosition);
+        }
```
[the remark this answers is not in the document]
tomekzaw: Looks good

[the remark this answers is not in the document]
bernhardoj: Updated

tomekzaw (approved):
@bernhardoj Thanks for this PR as well as for applying all code review changes. The code looks good to me now.

Skalakid (approved):
LGTM

PR #558, opened 2024-12-04
Move parsing logic to separate class on iOS

tomekzaw:
### Details
This PR moves parsing logic involving worklet runtime into a separate class called `MarkdownParser` on iOS as well as adds some memoization to avoid unnecessary work.

### Related Issues

GH_LINK

### DIFF
--- a/apple/MarkdownParser.mm
+++ b/apple/MarkdownParser.mm
@@ -1,0 +1,68 @@
+#import "MarkdownParser.h"
+#import <RNLiveMarkdown/MarkdownGlobal.h>
+#import <React/RCTLog.h>
+#import <mutex>
+
+@implementation MarkdownParser {
+  NSString *_prevText;
+  NSNumber *_prevParserId;
+  NSArray<MarkdownRange *> *_prevMarkdownRanges;
+}
+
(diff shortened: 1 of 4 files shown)


apple/MarkdownParser.mm
```diff
@@ -0,0 +1,68 @@
...
+#import <mutex>
+
+@implementation MarkdownParser {
+  NSString *_prevText;
+  NSNumber *_prevParserId;
+  NSArray<MarkdownRange *> *_prevMarkdownRanges;
+}
+
+- (NSArray<MarkdownRange *> *)parse:(NSString *)text withParserId:(nonnull NSNumber *)parserId {
+  @synchronized (self) {
```
j-piasecki, line 13: Is `@synchronized` needed here? Or do you just want to be explicit about the behavior?

tomekzaw: `MarkdownParser` can be used from multiple threads (i.e. UI and JS thread) and we read/write `_prev*` fields so that's why I added `@synchronized`.

tomekzaw: Yep, moving the lock would make `@synchronized` unnecessary. But the lock itself is also subject to change as I'd like to eliminate keeping worklet runtime in global at some point so probably the code in the current shape is more future-proof.

j-piasecki (approved):

PR #559, opened 2024-12-04
Introduce `MarkdownRange` class on Android

tomekzaw:
### Details
This PR adds an intermediate step in formatting `applyMarkdownFormatting` method on Android. Instead of using `JSONObject` obtained from parser directly, first we populate `markdownRanges` list consisting of `MarkdownRange` objects. Then, we call `applyRange` while iterating over the array.

This change will make it simpler to split the implementation into smaller parts and memoize the parser output as Java objects instead of memoizing JSON string.

### Related Issues

GH_LINK


j-piasecki (approved):
Type could be an enum, though I guess it would require more changes to the code.

PR #568, opened 2024-12-07
Skip formatting if data unchanged on Android

tomekzaw:
### Details
This PR skips removing and applying spans on Android if the text, ranges and styles are unchanged to previous run of `MarkdownFormatter` as span operations are relatively time-consuming on Android.

In particular, this lets us save some time during mounting phase as per systraces below:

#### Before
Screenshot 2024-12-07 at 12 04 14

#### After
Screenshot 2024-12-07 at 12 03 13

### Related Issues

GH_LINK

### DIFF
--- a/android/src/main/java/com/expensify/livemarkdown/MarkdownFormatter.java
+++ b/android/src/main/java/com/expensify/livemarkdown/MarkdownFormatter.java
@@ -15,6 +15,9 @@ import java.util.Objects;
 
 public class MarkdownFormatter {
   private final @NonNull AssetManager mAssetManager;
+  private String mPrevText;
+  private List<MarkdownRange> mPrevMarkdownRanges;
+  private MarkdownStyle mPrevMarkdownStyle;
 
   public MarkdownFormatter(@NonNull AssetManager assetManager) {
     mAssetManager = assetManager;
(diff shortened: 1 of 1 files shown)


tomekzaw:
Note: I feel like there could be some edge-cases where such optimization leads to incorrect result but I wasn't able to find one so far. This needs a bit of thought prior to merging.

@289Adam289 Could you please help me test this PR for regressions? Thanks in advance.

289Adam289:
Sure, happy to help. I will get to it soon

android/src/main/java/com/expensify/livemarkdown/MarkdownFormatter.java
```diff
@@ -24,8 +27,19 @@ public void format(SpannableStringBuilder ssb, List<MarkdownRange> markdownRange
     try {
       Systrace.beginSection(0, "format");
       Objects.requireNonNull(markdownStyle, "mMarkdownStyle is null");
+
+      String text = ssb.toString();
+      if (text.equals(mPrevText) && markdownRanges == mPrevMarkdownRanges && markdownStyle == mPrevMarkdownStyle) {
+        // Use shallow comparison of markdown ranges and markdown style
+        // to optimistically skip removing and applying the same spans
+        return;
+      }
```
j-piasecki, line 36: Isn't there anything that could alter the styles in between format passes? I'm not sure how it will behave with the commit hook when the text doesn't change but the node is cloned.

289Adam289:
BUG:
Coping and pasting text on top of exactly the same text formatted range breaks markdown formatting
Reproduction steps:
1. Write `*world*`
2. Copy exactly `*world*`
3. select `*world*`
4. Paste

https://github.com/user-attachments/assets/96518f2c-705c-4dea-ab8a-0a33a3c35aeb

tomekzaw:
Thanks @289Adam289 for testing out this PR and finding the regressions.

As for https://github.com/Expensify/react-native-live-markdown/pull/568#issuecomment-2536202684, right now I don't have any idea on how to modify the logic to avoid this bug.

PR #578, opened 2024-12-10
add mock

s77rt:
cc @tomekzaw 

### Related Issues

https://github.com/Expensify/App/issues/52475

### DIFF
--- a/src/MarkdownTextInput.tsx
+++ b/src/MarkdownTextInput.tsx
@@ -28,10 +28,9 @@ function initializeLiveMarkdownIfNeeded() {
   if (initialized) {
     return;
   }
-  if (!NativeLiveMarkdownModule) {
-    throw new Error('[react-native-live-markdown] NativeLiveMarkdownModule is not available');
+  if (NativeLiveMarkdownModule) {
+    NativeLiveMarkdownModule.install();
   }
-  NativeLiveMarkdownModule.install();
   if (!global.jsi_setMarkdownRuntime) {
     throw new Error('[react-native-live-markdown] global.jsi_setMarkdownRuntime is not available');
   }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -26,6 +26,6 @@
     "verbatimModuleSyntax": true,
     "typeRoots": ["node_modules/@types"]
   },
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "mock/**/*"],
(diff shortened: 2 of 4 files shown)


src/MarkdownTextInput.tsx
```diff
@@ -28,10 +28,9 @@ function initializeLiveMarkdownIfNeeded() {
   if (initialized) {
     return;
   }
-  if (!NativeLiveMarkdownModule) {
-    throw new Error('[react-native-live-markdown] NativeLiveMarkdownModule is not available');
+  if (NativeLiveMarkdownModule) {
+    NativeLiveMarkdownModule.install();
```
tomekzaw, line 32: What's the reason for this change? If `NativeLiveMarkdownModule` is not injected properly, we'd like to fail immediately instead of showing a non-actionable error "global.jsi_setMarkdownRuntime is not available" that sounds more like an actual implementation bug rather than setup issue.

tsconfig.json
```diff
@@ -26,6 +26,6 @@
     "verbatimModuleSyntax": true,
     "typeRoots": ["node_modules/@types"]
   },
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "mock/**/*"],
```
tomekzaw, line 29: Can we move `mock` directory to `src/` directory? Or would this affect the paths in imports? Let's avoid `@expensify/react-native-live-markdown/src/mock` if possible (with `src/mock`)

src/MarkdownTextInput.tsx
s77rt: This is required for the mock to work. Otherwise we'd have to [mock react-native TurboModuleRegistry](https://github.com/Expensify/App/blob/4a9bd7cb3bbbbd95ec7b9a229611786ebe4dbbe8/__mocks__/react-native.ts#L45-L59).

tsconfig.json
s77rt: It will affect the import path indeed. It's better to keep the path as `@expensify/react-native-live-markdown/mock`

Kicu (approved):
I think the mock itself looks good.

I'm not sure about the `NativeLiveMarkdownModule.install()` part because I don't have much experience interacting with `TurboModuleRegistry`.
Everything else seems reasonable to me 👍

tomekzaw (approved):
LGTM, also tested all example apps and everything looks good. Thanks for the PR!

PR #582, opened 2024-12-11
Add zero-config web support

### LINKED ISSUE #579: Use `.native.tsx` instead of `.web.tsx`
React web apps don't use the `.web.tsx` pattern. However, all native apps do use platform extensions.

I'd propose changing `MarkdownTextInput.tsx` to have the web code, and `MarkdownTextInput.tsx` to have the React Native code. This way, Web developers don't need additional config to install the library.

I'm happy to do this in a PR if you'd merge it.

nandorojo:
### Details

(Still a draft)

Removes any dependency on React Native, React Native Web, for Web users. Nothing changes for native. This should drastically fix any installation steps / config issues for Web users. Without these changes, web users will flood your issues constantly lol.

- [ ] Rename `.web.tsx` files to be `.native.tsx`
- [ ] Remove usages of inline `Platform.select` in favor of `.native.tsx`
- [ ] Use web-specific code as the main file for any cross-platform files.
- [ ] Remove `StyleSheet.create`

### Related Issues

GH_LINK

Closes #579


nandorojo:
Some feedback:

Currently `webStyleUtils` uses the following from React Native Web:

1. `createReactDOMStyle` [(code)](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/StyleSheet/compiler/createReactDOMStyle.js)
  a. This adds support for a bunch of stuff that's basically never used, such as `marginInline`
  b. I'm thinking this could be simplified and implemented directly in library code, especially as React Native starts to support web code
  c. The only one that we _might_ want to keep is the `flex` styles. I think this should be implemented by this lib instead of importing from RNW.
2. `preprocessStyle` [(code)](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/StyleSheet/preprocess.js#L109-L220)
  a. The only real function this modules serves is to give deprecation warnings about the styles. I think this step should also get removed, as React Native now supports (and prefers) the web styles.
3. `dangerousStyleValue` [(code)](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/modules/setValueForStyles/dangerousStyleValue.js)
 a. This is nothing more than checking for nullish values / numbers / empty strings. This should be upstreamed to this library.

Tamagui, Gluestack, etc have all upstreamed these functions into their own libs, [as seen here.](https://grep.app/search?q=createReactDOMStyle)

I haven't implemented these changes yet, I'll wait for feedback first. CC @tomekzaw

natew:
I can vouch for these changes as being beneficial generally for web compat. Ideally it'd be a mostly headless setup and all native files are in .native, with web being the "default". This makes for no-config setups.

Issue #584, opened 2024-12-12
Custom parser duplicates the input content on web


Skalakid:
### Description:
After applying custom parser function to the Live Markdown Input, its content breaks. 

For example, after applying following parser:
```javaScript
function parser(input) {
  'worklet';

  const ranges = [];
  // Regular expressions for hashtags and mentions
  const hashtagRegex = /#(\w+)/g;
  const mentionRegex = /@(\w+)/g;

  // Find and push hashtag ranges
  let match;
  while ((match = hashtagRegex.exec(input)) !== null) {
    ranges.push({
      start: match.index,
      length: match[0].length,
      type: 'link', // type 'link' for hashtags
    });
  }

  // Find and push mention ranges
  while ((match = mentionRegex.exec(input)) !== null) {
    ranges.push({
      start: match.index,
      length: match[0].length,
      type: 'mention-user', // type 'mention-user' for user mentions
    });
  }

  return ranges;
}
```

https://github.com/user-attachments/assets/6dc07a05-5c55-4e8e-9d35-5584bf1f65da

### Reproduction steps
1. Open the example app
2. Write following text: `@a #1`

Skalakid:
Fixed in https://github.com/Expensify/react-native-live-markdown/pull/585

PR #585, opened 2024-12-12
Fix custom parsing on the web

Skalakid:
### Details

This PR adds range sorting in web range to HTML parser in `parserUtils.ts`. Web parsers requires sorted ranges to work properly, so we with fix we enforce the correct order of items

https://github.com/user-attachments/assets/ae384c30-a387-4cc4-a9b2-1b0eaafdcfed

### Related Issues

https://github.com/Expensify/react-native-live-markdown/issues/584

### Manual Tests

1. Open the example app
2. Write following text: `@a #1`

### DIFF
--- a/src/parseExpensiMark.ts
+++ b/src/parseExpensiMark.ts
@@ -292,4 +292,4 @@ function parseExpensiMark(markdown: string): MarkdownRange[] {
   return groupedRanges;
 }
 
-export default parseExpensiMark;
+export {parseExpensiMark, sortRanges};
--- a/src/web/utils/parserUtils.ts
+++ b/src/web/utils/parserUtils.ts
@@ -6,6 +6,7 @@ import {getCurrentCursorPosition, moveCursorToEnd, setCursorPosition} from './cu
 import {addStyleToBlock, extendBlockStructure, getFirstBlockMarkdownRange, isBlockMarkdownType} from './blockUtils';
 import type {InlineImagesInputProps, MarkdownRange} from '../../commonTypes';
 import {getAnimationCurrentTimes, updateAnimationsTime} from './animationUtils';
+import {sortRanges} from '../../parseExpensiMark';
 
 type Paragraph = {
   text: string;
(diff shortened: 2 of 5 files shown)


src/parseExpensiMark.ts
```diff
@@ -292,4 +292,4 @@ function parseExpensiMark(markdown: string): MarkdownRange[] {
   return groupedRanges;
 }
 
-export default parseExpensiMark;
+export {parseExpensiMark, sortRanges};
```
tomekzaw, line 295: I don't think it's a good idea to export and use `sortRanges` from `parseExpensiMark`. This is a self-contained parser used as default which we plan to remove from the repo at some point so we shouldn't use any of its pieces in the library code. Instead, let's extract `sortRanges` to a separate file (maybe something like `parserUtils.ts`?) and import it here from there.

tomekzaw (changes requested):

src/parseExpensiMark.ts
Skalakid: Okay, I moved them to the `rangeUtils` since `parserUtils.ts` already exists on the web

src/web/utils/parserUtils.ts
```diff
@@ -167,7 +168,8 @@ function parseRangesToHTMLNodes(
     return {dom: rootElement, tree: rootNode};
   }
 
-  const markdownRanges = ungroupRanges(ranges);
+  const sortedRanges = sortRanges(ranges);
+  const markdownRanges = ungroupRanges(sortedRanges);
```
tomekzaw, line 172: Why don't we move `ungroupRanges` to `rangeUtils.ts` as well?

src/web/utils/parserUtils.ts
```diff
@@ -167,7 +168,8 @@ function parseRangesToHTMLNodes(
     return {dom: rootElement, tree: rootNode};
   }
 
-  const markdownRanges = ungroupRanges(ranges);
+  const sortedRanges = sortRanges(ranges);
```
BartoszGrajdek, line 171: Can you add a comment here explaining either why we are doing this, or just what we're doing?

Skalakid: Done in [0d710a0](https://github.com/Expensify/react-native-live-markdown/pull/585/commits/0d710a0dcd7577aec7fd492aec3b8dd8abfdb117)

src/web/utils/parserUtils.ts
Skalakid: Done in [0d710a0](https://github.com/Expensify/react-native-live-markdown/pull/585/commits/0d710a0dcd7577aec7fd492aec3b8dd8abfdb117)

BartoszGrajdek (approved):
LGTM

Skalakid:
@tomekzaw bump

tomekzaw (approved):

Issue #588, opened 2024-12-16
'worklets/WorkletRuntime/WorkletRuntime.h' file not found when building for iOS with `USE_FRAMEWORKS=static`


tomekzaw:
0.1.207 + iOS

```
'worklets/WorkletRuntime/WorkletRuntime.h' file not found
```

![](https://github.com/user-attachments/assets/1668808e-bc18-4c12-bfd9-8bd0fc6ed8ed)

tomekzaw:
Workaround:

```diff
-#include <worklets/WorkletRuntime/WorkletRuntime.h>
+#include "WorkletRuntime.h"
```

```diff
-#include <worklets/SharedItems/Shareables.h>
+#include "Shareables.h"
```

tomekzaw:
Should be fixed with https://github.com/Expensify/react-native-live-markdown/pull/589

PR #589, opened 2024-12-17
Fix build with `USE_FRAMEWORKS=static`

tomekzaw:
### Details
Fixes https://github.com/Expensify/react-native-live-markdown/issues/588.
Fixes https://github.com/Expensify/App/issues/54256.

This PR fixes failing iOS build with `USE_FRAMEWORKS=static` caused by two problems.

Screenshot 2024-12-17 at 19 02 53

1. `'worklets/WorkletRuntime/WorkletRuntime.h' file not found` – this is caused by incorrect header structure in react-native-reanimated when `USE_FRAMEWORKS=static` is enabled. This is related to https://github.com/software-mansion/react-native-reanimated/pull/6614. The workaround here is to find path to `node_modules/react-native-reanimated`, append it with `apple/` or `Common/cpp/` and add them to header search paths. This needs to be fixed in react-native-reanimated; once fixed there we can remove the workaround here.

2. `react/renderer/consistency/ShadowTreeRevisionConsistencyManager.h` file not found` – this is resolved by declaring a dependency on `React-rendererconsistency` podspec when `USE_FRAMEWORKS=static` is enabled. We already do the same thing for `React-FabricComponents`.

### Related Issues

GH_LINK

### DIFF
--- a/RNLiveMarkdown.podspec
+++ b/RNLiveMarkdown.podspec
@@ -4,6 +4,10 @@ react_native_node_modules_dir = ENV['REACT_NATIVE_NODE_MODULES_DIR'] || File.joi
 react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir, 'react-native/package.json')))
 react_native_minor_version = react_native_json['version'].split('.')[1].to_i
 
+pods_root = Pod::Config.instance.project_pods_root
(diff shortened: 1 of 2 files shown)


luacmartins (approved):

tomekzaw:
Thanks @luacmartins for approving this!

PR #591, opened 2024-12-23
Prevent italic or strikethrough emojis on Android

Skalakid:
### Details

This PR fixes adding italics and strikethrough style to the emojis on Android and web platforms. It adds range splitting algorithm in TypeScript to `parseExpensiMark` function so we won't interfere with the results of the custom parsers. This solution replaces https://github.com/Expensify/react-native-live-markdown/pull/534 PR, as discussed internally cc @tomekzaw 
### Related Issues

### Manual Tests

1. To the text that contains emojis, add italics styling, e.g. `_🥰 test 😇 test 🥹_`
2. Verify if the text has italics style and emojis don't
3. Verify if it still works when adding other styles around it, e.g. `# *~_🥰 test 😇 test 🥹_~*`
4. Verify if emojis aren't strikethrough
5. Test it with different style combinations

### DIFF
--- a/src/__tests__/splitRangesOnEmojis.test.ts
+++ b/src/__tests__/splitRangesOnEmojis.test.ts
@@ -1,0 +1,64 @@
+import type {MarkdownRange} from '../commonTypes';
+import {splitRangesOnEmojis} from '../rangeUtils';
+
+test('No Overlap', () => {
+  const markdownRanges: MarkdownRange[] = [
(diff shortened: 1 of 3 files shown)


tomekzaw (dismissed):
Thanks for this PR! I know it's been a long journey but the final effect is definitely worth it.

tomekzaw (approved):

PR #597, opened 2025-01-11
Stabilised the output of splitRangesOnEmojis

shubham1206agra:
### Details

This PR stabilised the output of splitRangesOnEmojis (made it sorted) since the input should be sorted to make this function work properly.

### Related Issues

https://github.com/Expensify/App/issues/55115

### Manual Tests

1. Open composer / markdown input.
2. Try to add `_*😊*_` in the input.
3. Make sure emoji is formatted, and input remains the same.

### DIFF
--- a/src/__tests__/splitRangesOnEmojis.test.ts
+++ b/src/__tests__/splitRangesOnEmojis.test.ts
@@ -152,6 +152,36 @@ describe('multiple overlaps', () => {
     ]);
   });
 
+  test('splitting on one type back to back', () => {
+    let markdownRanges: MarkdownRange[] = [
+      {type: 'italic', start: 0, length: 20},
+      {type: 'strikethrough', start: 2, length: 12},
+      {type: 'emoji', start: 3, length: 1},
+      {type: 'emoji', start: 8, length: 2},
+      {type: 'strikethrough', start: 16, length: 5},
+      {type: 'emoji', start: 17, length: 1},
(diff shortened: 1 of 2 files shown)


Skalakid:
Hi @shubham1206agra thank you for the PR! Can you please provide more details about the issue you are fixing here? I don't see any proposal in the issues that you linked and there is no manual tests in the description

Skalakid:
Okay, thanks. In my opinion, your solution adds another layer of complexity to our Expensify default parser. It will be harder to maintain this code. Have you conducted any reliable performance tests on how much queue logic improves splitting logic? Executing another `sortRanges` function doesn't change the time complexity at all, O(nlogn + nlogn)  is still O(nlogn). So we are okay with another `sortRanges` to the code.

Another thing is that your solution doesn't fully fix the problem. Syntaxes are still tripled when you write `_*asd😭asd*_` or `_*'asd😭asd'*_` (where ' is equal to `):

https://github.com/user-attachments/assets/ffa3cf32-b7ed-4985-9511-ac207587b6fe

Skalakid:
Here is the [PR](https://github.com/Expensify/react-native-live-markdown/pull/598) with the temporary fix. As I wrote under the issue, feel free to change your solution and enhance how the web parser builds the HTML structure inside the Live Markdown. The solution will not only fix all edge cases connected to splitting ranges on emojis but also make the web implementation more bug-proof when working with custom parsers

PR #614, opened 2025-02-07
Implement format toggle feature

QichenZhu:
### Details

This PR implements a new feature to toggle bold and italic formatting using keyboard shortcuts or the context menu.

### Related Issues

https://github.com/Expensify/App/issues/55088
PROPOSAL: https://github.com/Expensify/App/issues/55088#issuecomment-2639014812

### Manual Tests

1. Open the web example app.
2. Clear the text box.
3. Enter the data and select the text according to the second column of the table below.
4. Press Command+B to toggle bold formatting, or press Command+I to toggle italic formatting.
5. Verify that the result matches the third column.

| Command | Before formatting | After formatting |
| -- | -- | -- |
| `formatBold` | Image | Image |
| `formatBold` | Image | Image |
| `formatBold` | Image | Image |
| `formatItalic` | Image | Image |
| `formatItalic` | Image | Image |
| `formatItalic` | Image | Image |
| `formatBold` | Image | Image |
| `formatBold` | Image | Image |
| `formatItalic` | Image | Image |
| `formatItalic` | Image | Image |

https://github.com/user-attachments/assets/98009372-4637-435b-990d-9da5eb0a6c31

### DIFF
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -24,6 +24,7 @@ import {getElementHeight, getPlaceholderValue, isEventComposing, normalizeValue,
 import {parseToReactDOMStyle, processMarkdownStyle} from './web/utils/webStyleUtils';
 import {forceRefreshAllImages} from './web/inputElements/inlineImage';
 import type {MarkdownRange, InlineImagesInputProps} from './commonTypes';
+import {removeFormat} from './web/utils/formatToggleUtils';
 
 const useClientEffect = typeof window === 'undefined' ? useEffect : useLayoutEffect;
 
@@ -247,6 +248,12 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
 
         if (!formatSelection) {
           return parseText(parser, target, parsedText, processedMarkdownStyle, cursorPosition);
+        }
+
+        // Undo the formatting if the selection is already in the given format.
+        const formatRemovalResult = removeFormat(parsedText, contentSelection.current.start, contentSelection.current.end, formatCommand);
+        if (formatRemovalResult) {
+          return parseText(parser, target, formatRemovalResult.updatedText, processedMarkdownStyle, cursorPosition + formatRemovalResult.cursorOffset, true);
         }
 
         const selectedText = parsedText.slice(contentSelection.current.start, contentSelection.current.end);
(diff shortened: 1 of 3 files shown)


tomekzaw:
Hi @QichenZhu, thanks for this PR! I'm astonished by the quality of the PR description, in particular the table with manual tests, really nice!

As for the implementation, I'm not really convinced if it's a good idea to have all the logic responsible for removing formatting in `react-native-live-markdown`. As you've mentioned, we reuse some of the regexps from expensify-common which is not advisable.

When we discussed this with @BartoszGrajdek and @Skalakid earlier, we thought that `MarkdownTextInput` would only expose a callback that would be passed from app logic, i.e. Expensify/App. This way `react-native-live-markdown` can be still syntax-agnostic to some extend as the developer can parametrize the behavior from outside. What do you think about this approach?

QichenZhu:
@tomekzaw Thanks for your review.

Yeah, it makes more sense to just expose callbacks.

We currently have a callback `formatSelection`. Do you prefer to add a separate callback like `unformatSelection`, or to update the interface of `formatSelection` to support toggling?

QichenZhu:
Hi @tomekzaw, I drafted another PR https://github.com/Expensify/react-native-live-markdown/pull/615. Could you check if it aligns closer with your thoughts? I'm open to any feedback. Thanks!

tomekzaw:
@QichenZhu Thanks! Yep, this is a bit closer, but I don't see the logic responsible for removing the formatting if the selected text is wrapped in bold or italic, how does it work then?

QichenZhu:
> This logic will be on E/App side, right?

Yes, that's my plan.

> In that case, maybe we'll update the logic in example app so toggling also works here?

Done. However, is it a good idea to duplicate the code in both E/App and the example? It will add maintenance effort.

tomekzaw:
> Yes, that's my plan.

Good, thanks!

> Done. However, is it a good idea to duplicate the code in both E/App and the example? It will add maintenance effort.

I agree that this is additional maintenance effort, but actually there's no need to be extra consistent between E/App and live-markdown, if the implementation in live-markdown will be outdated, it's not a big problem.

Alternatively, we can move the implementation to `expensify-common` and import it in both places, what do you think?

QichenZhu:
Could we close this in favor of https://github.com/Expensify/react-native-live-markdown/pull/615?

PR #617, opened 2025-02-10
Fix setting border radius on web in singleline mode

Skalakid:
### Details

This PR fixes adding border radius style to the markdowns in the single line input mode.

### Related Issues

GH_LINK

### Manual Tests

1. Add, for example
```
      mentionUser: {
        borderRadius: 5,
      },
```
 to the `markdownStyles`
2. Change the MarkdownTextInput component to the single line mode by deactivating the `multiline` flag
3. Verify if the border radius style was applied

### DIFF
--- a/src/MarkdownTextInputDecoratorViewNativeComponent.ts
+++ b/src/MarkdownTextInputDecoratorViewNativeComponent.ts
@@ -37,14 +37,17 @@ interface MarkdownStyle {
   mentionHere: {
     color: ColorValue;
     backgroundColor: ColorValue;
+    borderRadius?: Float;
   };
   mentionUser: {
     color: ColorValue;
     backgroundColor: ColorValue;
+    borderRadius?: Float;
   };
   mentionReport: {
     color: ColorValue;
     backgroundColor: ColorValue;
+    borderRadius?: Float;
   };
   inlineImage: {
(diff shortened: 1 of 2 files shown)


tomekzaw:
We don't need to add it to native right now, we can do that later.

src/MarkdownTextInputDecoratorViewNativeComponent.ts
BartoszGrajdek, line 1: I'm not entirely sure how we should handle these types. It can mislead developers into thinking that we support `borderRadius` on native devices. Maybe we should make a separate type for `MarkdownStyle` on web that will be a union of this interface and a new type solely for the web?

BartoszGrajdek: In the past, we discussed changing web styles so that only the ones we support would be accepted. This shouldn't be too hard to implement; I'm just not sure if that's exactly what we're aiming for.

tomekzaw: I don't think that having `borderRadius` in markdown style also on native is a problem. We can just leave it as a no-op for now.

BartoszGrajdek (approved):
LGTM!

PR #626, opened 2025-02-27
Fix inline images in multiple Live Markdown inputs

Skalakid:
### Details

This PR fixes the problem with infinite loading indicators when two Live Markdown inputs with the same content are present. Also it changes the way how we calculate scaled image height to fix problems related to handling multiple inputs.

### Related Issues

https://github.com/Expensify/App/issues/56739

### Manual Tests

1. Add multiple live markdown inputs that have the same input value with inline image markdown
2. Refresh the page
3. Verify if inline images are loaded in both inputs


BartoszGrajdek (approved):
LGTM!

Issue #602, opened 2025-01-28
Linking error `error: undefined symbol: worklets::extractWorkletRuntime` on Android


tomekzaw:
```
Execution failed for task ':expensify_react-native-live-markdown:buildCMakeDebug[arm64-v8a]'.
> com.android.ide.common.process.ProcessException: ninja: Entering directory `/Users/foo/expensifyForks/swmlabs/expensify-app-fork/node_modules/@expensify/react-native-live-markdown/android/.cxx/Debug/2er395t6/arm64-v8a'
  [0/2] Re-checking globbed directories...
  [1/1] Linking CXX shared library /Users/foo/expensifyForks/swmlabs/expensify-app-fork/node_modules/@expensify/react-native-live-markdown/android/build/intermediates/cxx/Debug/2er395t6/obj/arm64-v8a/liblivemarkdown.so
  FAILED: /Users/foo/expensifyForks/swmlabs/expensify-app-fork/node_modules/@expensify/react-native-live-markdown/android/build/intermediates/cxx/Debug/2er395t6/obj/arm64-v8a/liblivemarkdown.so 
[...]

PR #627, opened 2025-02-28
Fix linking error `error: undefined symbol: worklets::extractWorkletRuntime` on Android

tomekzaw:
Co-authored-by: Łukasz Kosmaty <lukasz.kosmaty@swmansion.com>

### Details
This PR fixes building RN apps with `@expensify/react-native-live-markdown` on Android after Gradle Sync.

Big kudos to @lukmccall for investigating this issue and coming up with a clever fix!

### Related Issues
Fixes https://github.com/Expensify/react-native-live-markdown/issues/602.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,6 +9,10 @@ buildscript {
   }
 }
 
+task prepareKotlinBuildScriptModel {
+  // This task is run during Gradle Sync in Android Studio.
+}
+
 def reactNativeArchitectures() {
   def value = project.getProperties().get("reactNativeArchitectures")
   return value ? value.split(",") : ["armeabi-v7a", "x86", "x86_64", "arm64-v8a"]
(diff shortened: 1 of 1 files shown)


j-piasecki (approved):
Is this something we would like to mention in the Reanimated docs?

android/build.gradle
```diff
@@ -9,6 +9,10 @@ buildscript {
   }
 }
 
+task prepareKotlinBuildScriptModel {
```
WoLewicki, line 12: why is this task added explicitly here?

tomekzaw: Gradle Sync executes `:prepareKotlinBuildScriptModel` which runs all `prepareKotlinBuildScriptModel` tasks in a project. Often there's only `:app:prepareKotlinBuildScriptModel` but here we register an empty task `:expensify_react-native-live-markdown:prepareKotlinBuildScriptModel` so it runs during Gradle Sync.

WoLewicki (approved):

PR #566, opened 2024-12-06
Replace the commit hook with a wrapper component based on `display: contents`

j-piasecki:
### Details
React Native 0.77 comes with support for `display: contents` which would allow to make a wrapper component that doesn't impact layout in any way. This can be used to replace the commit hook to apply markdown styles without cloning the tree.

### Related Issues

GH_LINK

PR #634, opened 2025-03-07
Remove old architecture support

tomekzaw:
### Details

This PR removes support for the Old Architecture for Android and iOS.

### Motivation

The current implementation of `@expensify/react-native-live-markdown` is mostly good but also overly complex. It makes it difficult to maintain code and resolve current issues in its current shape. We don't have enough time to maintain both architectures at once. Considering that New Expensify app successfully runs with the New Architecture, we can phase out the old arch support and focus on existing issues. There's a PR https://github.com/Expensify/react-native-live-markdown/pull/566 from @j-piasecki that refactors the RN layer and uses `display: contents;` that is available only on the New Architecture.

https://swmansion.slack.com/archives/C04878MDF34/p1741365120700689

### Related Issues

GH_LINK


android/build.gradle
```diff
@@ -190,12 +177,10 @@ dependencies {
...
-  react {
-    jsRootDir = file("../src/")
-    libraryName = "LiveMarkdown"
-    codegenJavaPackageName = "com.expensify.livemarkdown"
-  }
+react {
+  jsRootDir = file("../src/")
+  libraryName = "LiveMarkdown"
+  codegenJavaPackageName = "com.expensify.livemarkdown"
 }
```
j-piasecki, line 184: This shouldn't be neccessary, due to `codegenConfig` field in `package.json`

apple/MarkdownTextInputDecoratorView.mm
```diff
@@ -6,59 +6,34 @@
...
   __weak UIView<RCTBackedTextInputViewProtocol> *_backedTextInputView;
   __weak RCTBackedTextFieldDelegateAdapter *_adapter;
   __weak RCTUITextView *_textView;
 }
 
 - (void)didMoveToWindow {
-#ifdef RCT_NEW_ARCH_ENABLED
   if (self.superview.superview == nil) {
     return;
   }
```
j-piasecki, line 26: It should be possible to merge Decorator view and Decorator component view now (in a follow-up, of course)

j-piasecki (dismissed):

android/build.gradle
tomekzaw: What exactly do you suggest? Should I remove only `codegenJavaPackageName` or something more?

apple/MarkdownTextInputDecoratorView.mm
tomekzaw: What exact changes do you suggest?

android/build.gradle
j-piasecki: I believe entire `react` block isn't needed

apple/MarkdownTextInputDecoratorView.mm
j-piasecki: I thought that `MarkdownTextInputDecoratorComponentView` was a wrapper around `MarkdownTextInputDecoratorView` but now that I've checked it looks like two independent implementations.

Is `MarkdownTextInputDecoratorView` used at all with these changes?

android/build.gradle
tomekzaw: Cool, I'll remove that, we still need `apply plugin: "com.facebook.react"` though, right?

apple/MarkdownTextInputDecoratorView.mm
tomekzaw: Let me check that

android/build.gradle
tomekzaw: Removed in 8b3d88b.

tomekzaw: Removed in 8b3d88b.

j-piasecki: Can you check if updating `codegenConfig` to
```
  "codegenConfig": {
    "name": "RNLiveMarkdownSpec",
    "type": "all",
    "jsSrcsDir": "src",
    "android": {
      "javaPackageName": "com.expensify.livemarkdown"
    }
  }
```
helps?

tomekzaw: That helped. Changed in 1f60bf2.

j-piasecki (approved):

PR #656, opened 2025-04-03
[WEB] Fix cursor moves several characters when deleting markdowns on multiline message 

VickyStash:
### Details

Fix cursor moves several characters on android web platform when deleting markdowns on multiline message.

### Related Issues

https://github.com/Expensify/App/issues/50687
https://github.com/Expensify/App/issues/57024
Proposal: https://github.com/Expensify/App/issues/50687#issuecomment-2484120140

### Manual Tests

1. Enter a bold text in multiline.
2. Delete * at the end of the text.
3. Make sure the cursor position is expected.
4. Experiment with typing quickly when there are a lot of markdown styles inside the input to make sure there are no side affects due to updates.

<details>
<summary>WEB</summary>

https://github.com/user-attachments/assets/43b15d84-5783-41f1-8aa2-969bd732cd90

</details>

<details>
<summary>WEB iOS</summary>

https://github.com/user-attachments/assets/0fe664cf-02d6-4583-8137-2305023ffc30

</details>

<details>
<summary>WEB Android</summary>

https://github.com/user-attachments/assets/97d21bad-9fcc-4cd7-a2a0-39dea99243a7

</details>

### Linked PRs

N/A


WebExample/__tests__/input.spec.ts
```diff
@@ -12,17 +12,19 @@ test.describe('typing', () => {
...
 
     await inputLocator.focus();
-    await inputLocator.pressSequentially(TEST_CONST.EXAMPLE_CONTENT);
+    await inputLocator.pressSequentially(TEST_CONST.EXAMPLE_CONTENT, {delay: TEST_CONST.USER_TYPING_DELAY});
 
     expect(await getElementValue(inputLocator)).toEqual(TEST_CONST.EXAMPLE_CONTENT);
   });
 
   test('fast type cursor position', async ({page}) => {
+    test.setTimeout(60000);
```
Skalakid, line 21: Why do we need this?

Skalakid (commented):
I see that these changes break quick typing a bit, since now we need to pass a slight delay to our tests. I'm worried that on slower the required delay will increase, and some bugs may appear :/
Also, there are some more places in our library where the input structure is modified, for example, the `insertText` function in the main web component. Maybe we also need to add `runAfterInteractions` there. Can we check if we can remove this delay in any way?

WebExample/__tests__/input.spec.ts
VickyStash: The default test timeout is 30sec, with added [delay](https://github.com/Expensify/react-native-live-markdown/pull/656/files/0812c4b17ba7c718c7f6752b5d76364c45ec3e0b#diff-e938cac7d798d0eb0d40cb2619dbb2693a56e1b04508cb04fa05cb4c03d9dc08R27) this test takes ~40sec to run, so I needed to increase the timeout.

tomekzaw: Can we use fake timers instead? https://jestjs.io/docs/timer-mocks

VickyStash:
> I see that these changes break quick typing a bit, since now we need to pass a slight delay to our tests. I'm worried that on slower devices the required delay will increase, and some bugs may appear :/

The delay I've added is a simulation of human typing. But I understand you concern about slow-performant devices.

> Also, there are some more places in our library where the input structure is modified, for example, the insertText function in the main web component. Maybe we also need to add runAfterInteractions there. 

I'll check it!

> Can we check if we can remove this delay in any way?

I've tried to do it on Friday but had no luck unfortunately.

WebExample/__tests__/input.spec.ts
tomekzaw: Oops, you're right.

VickyStash:
> Also, there are some more places in our library where the input structure is modified, for example, the insertText function in the main web component. Maybe we also need to add runAfterInteractions there.

It looks like all of the places that update the input structure use the [updateInputStructure](https://github.com/Expensify/react-native-live-markdown/pull/656/files#diff-a59d7bddc330f4a9d78c820857f1a79469edf0b10f55ccf4b00b68d98940a5afL269) method anyway.

For example `insertText` calls `handleOnChangeText` => `handleOnChangeText` calls `parseText` or other methods that later calls `parseText` => `parseText` calls `updateInputStructure`.

I've also done some testing on the real low-performant device (Android - Redmi Note 7) and haven't had problems, but we def will need to keep an eye on it.

Skalakid (approved):
LGTM

Skalakid:
Hello @VickyStash, unfortunately, `InteractionManager.runAfterInteractions` causes [the following issue](https://github.com/Expensify/react-native-live-markdown/pull/661#issuecomment-2809812398) in Expensify NewDot App. I think we need to revert your changes, since it will be a deploy blocker

VickyStash:
> Hello @VickyStash, unfortunately, `InteractionManager.runAfterInteractions` causes [the following issue](https://github.com/Expensify/react-native-live-markdown/pull/661#issuecomment-2809812398) in Expensify NewDot App. I think we need to revert your changes, since it will be a deploy blocker

It feels like this PR was reverted by mistake, as the original issue [exists on E/App main](https://github.com/Expensify/react-native-live-markdown/pull/661#issuecomment-2811777751), but the updates of this PR never reached the main.
@Skalakid Have you tested the issue with/without these PR updates to confirm?

Skalakid:
Yes, I've tested the issue with and without your PR changes. I've also tested different versions of the Live Markdown, and only when `InteractionManager.runAfterInteractions` is added the issue appears. [Here](https://github.com/Expensify/react-native-live-markdown/pull/661#issuecomment-2815215903) is more info

PR #683, opened 2025-05-23
fix Firefox arrow navigation bug

jmusial:
Fixes navigation bug on Firefox, where navigation using arrows over emojis requires 2 key presses.

https://github.com/user-attachments/assets/0d559adb-0841-4fef-9ee1-9485dc86445f

### Details

In Firefox, the arrow key navigation treats surrogate pairs (like emojis) as two separate code units, while Chromium browsers treat them as a single character. This discrepancy causes the “double-arrow press” behaviour.

### Related Issues

GH_LINK

### Manual Tests

1. Open editor with some emojis typed in
2. Try navigating with your arrows
3. Results should be consistent across all browsers and 1 key press should move cursor over 1 emoji

#### Before 

https://github.com/user-attachments/assets/41ef5add-eee1-4ce6-ba3c-3b2c7c9f44fd

#### After 

https://github.com/user-attachments/assets/fa7b0793-d65c-4076-85c6-1e3bd8d49b29

### DIFF
--- a/src/web/utils/firefoxUtils.ts
+++ b/src/web/utils/firefoxUtils.ts
@@ -13,27 +13,35 @@ function handleFirefoxArrowKeyNavigation(target: MarkdownTextInputElement, isSel
   const text = target.value;
 
   const segmenter = new Intl.Segmenter('en', {granularity: 'grapheme'});
-  const graphemes = Array.from(segmenter.segment(text));
+  const segments = segmenter.segment(text);
 
   if (direction === 'right') {
     const cursorPos = currentSelection.end;
-    const nextGrapheme = graphemes.find(({index, segment}) => {
-      const segmentEnd = index + segment.length;
-      return cursorPos < segmentEnd;
-    });
+    let newCursorPos = text.length;
 
-    const newCursorPos = nextGrapheme ? nextGrapheme.index + nextGrapheme.segment.length : text.length;
+    // eslint-disable-next-line no-restricted-syntax
+    for (const {index, segment} of segments) {
+      const segmentEnd = index + segment.length;
+      if (cursorPos < segmentEnd) {
+        newCursorPos = segmentEnd;
+        break;
+      }
+    }
     setCursorPosition(target, isSelectionEvent ? currentSelection.start : newCursorPos, newCursorPos);
-  } else {
-    const cursorPos = currentSelection.start;
-    const prevGrapheme = graphemes.findLast(({index, segment}) => {
-      const segmentEnd = index + segment.length;
-      return segmentEnd < cursorPos;
-    });
-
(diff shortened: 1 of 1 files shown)


jmusial:
recheck

src/MarkdownTextInput.web.tsx
```diff
@@ -526,6 +528,15 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
           onKeyPress(event);
         }
 
+        // Handle ArrowRight for consistent navigation across grapheme clusters (like emojis) on firefox
```
tomekzaw, line 531:
```suggestion
        // Handle ArrowRight for consistent navigation across grapheme clusters (like emojis) on Firefox
```

src/MarkdownTextInput.web.tsx
```diff
@@ -526,6 +528,15 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
...
+        if (e.key === 'ArrowRight' && BrowserUtils.isFirefox && !nativeEvent.altKey) {
```
tomekzaw, line 532:
```suggestion
        if (BrowserUtils.isFirefox && e.key === 'ArrowRight' && !nativeEvent.altKey) {
```

src/web/utils/__tests__/firefoxUtils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+jest.mock('../cursorUtils', () => ({
+  ...jest.requireActual('../cursorUtils'),
+  getCurrentCursorPosition: jest.fn(),
+  setCursorPosition: jest.fn(),
+}));
+
+describe('handleFirefoxArrowKeyNavigation', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
```
tomekzaw, line 20: Let's add a newline here
```suggestion
  });

```

src/web/utils/firefoxUtils.ts
```diff
@@ -0,0 +1,26 @@
...
+  const currentSelection = getCurrentCursorPosition(target);
+  if (!currentSelection) {
+    return;
+  }
+
+  const text = target.value;
+  const cursorPos = currentSelection.end;
+
+  const segmenter = new Intl.Segmenter('en', {granularity: 'grapheme'});
+  const graphemes = Array.from(segmenter.segment(text));
```
tomekzaw, line 14: How long does this step take for longer values (approx 5000 characters)?

Can we somehow avoid allocating an array of graphemes and use the returned iterable instead?

Skalakid: Let's change it, it's always some optimization

PR #689, opened 2025-06-03
Fix navigation breaking markdown

j-piasecki:
### Related Issues

GH_LINK

### DIFF
--- a/apple/MarkdownTextInputDecoratorComponentView.mm
+++ b/apple/MarkdownTextInputDecoratorComponentView.mm
@@ -28,6 +28,7 @@ using namespace facebook::react;
   MarkdownTextFieldObserver *_markdownTextFieldObserver;
   __weak RCTUITextView *_textView;
   __weak RCTUITextField *_textField;
+  bool _observersAdded;
 }
 
 + (ComponentDescriptorProvider)componentDescriptorProvider
@@ -46,6 +47,7 @@ using namespace facebook::react;
   if (self = [super initWithFrame:frame]) {
     static const auto defaultProps = std::make_shared<const MarkdownTextInputDecoratorViewProps>();
     _props = defaultProps;
+    _observersAdded = false;
   }
 
   return self;
@@ -53,14 +55,36 @@ using namespace facebook::react;
 
 - (void)didAddSubview:(UIView *)subview
 {
-  react_native_assert([subview isKindOfClass:[RCTTextInputComponentView class]] && "Child component of MarkdownTextInputDecoratorComponentView is not an instance of RCTTextInputComponentView.");
-  RCTTextInputComponentView *textInputComponentView = (RCTTextInputComponentView *)subview;
-  UIView<RCTBackedTextInputViewProtocol> *backedTextInputView = [textInputComponentView valueForKey:@"_backedTextInputView"];
-
   _markdownUtils = [[RCTMarkdownUtils alloc] init];
   [_markdownUtils setMarkdownStyle:_markdownStyle];
   [_markdownUtils setParserId:_parserId];
 
+  [self addTextInputObservers:subview];
+}
+
+- (void)willMoveToWindow:(UIWindow *)newWindow
+{
+  if (newWindow != nil) {
+    if (self.subviews.count > 0) {
+      [self addTextInputObservers:self.subviews[0]];
+    }
+  } else {
+    [self removeTextInputObservers];
+  }
+}
+
+- (void)addTextInputObservers:(UIView *)textInput
+{
+  if (_observersAdded) {
+    return;
+  }
+  
+  react_native_assert([textInput isKindOfClass:[RCTTextInputComponentView class]] && "Child component of MarkdownTextInputDecoratorComponentView is not an instance of RCTTextInputComponentView.");
+  RCTTextInputComponentView *textInputComponentView = (RCTTextInputComponentView *)textInput;
+  UIView<RCTBackedTextInputViewProtocol> *backedTextInputView = [textInputComponentView valueForKey:@"_backedTextInputView"];
+ 
+  _observersAdded = true;
+  
   if ([backedTextInputView isKindOfClass:[RCTUITextField class]]) {
     _textField = (RCTUITextField *)backedTextInputView;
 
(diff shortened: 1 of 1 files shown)


apple/MarkdownTextInputDecoratorComponentView.mm
```diff
@@ -46,20 +47,35 @@ - (instancetype)initWithFrame:(CGRect)frame
   if (self = [super initWithFrame:frame]) {
     static const auto defaultProps = std::make_shared<const MarkdownTextInputDecoratorViewProps>();
     _props = defaultProps;
+    _observersAdded = false;
+    _markdownUtils = [[RCTMarkdownUtils alloc] init];
```
j-piasecki, line 51: Was there a reason this was originally recreated on every mount, or is creating it once safe?

war-in:
I checked it out in E/App and can confirm it tests well now 🚀 

https://github.com/user-attachments/assets/06ae83a7-1e60-4900-991d-04aa34b67e65

apple/MarkdownTextInputDecoratorComponentView.mm
tomekzaw: I just wanted to be extra safe and create a new instance of `RCTMarkdownUtils` each time the component is mounted to avoid using previous styles or parser in case something goes wrong. We can keep it your way if you want to.

apple/MarkdownTextInputDecoratorComponentView.mm
```diff
@@ -46,20 +47,35 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+}
 
-  _markdownUtils = [[RCTMarkdownUtils alloc] init];
-  [_markdownUtils setMarkdownStyle:_markdownStyle];
-  [_markdownUtils setParserId:_parserId];
+- (void)addTextInputObservers
+{
+  if (_observersAdded) {
+    return;
+  }
```
tomekzaw: Why can't we have `react_native_assert(!_observersAdded);` instead?

apple/MarkdownTextInputDecoratorComponentView.mm
```diff
@@ -115,11 +131,14 @@ - (void)didAddSubview:(UIView *)subview
   }
 }
 
-- (void)willMoveToWindow:(UIWindow *)newWindow
+- (void)removeTextInputObservers
 {
-  if (newWindow != nil) {
+  if (!_observersAdded) {
     return;
   }
```
tomekzaw: Why can't we have `react_native_assert(_observersAdded);` instead?

j-piasecki: We should be able to: [a3fbfa1](https://github.com/Expensify/react-native-live-markdown/pull/689/commits/a3fbfa1edc04e3edfa981c5f61fde466a75bbae9)

apple/MarkdownTextInputDecoratorComponentView.mm
j-piasecki: We should be able to: [a3fbfa1](https://github.com/Expensify/react-native-live-markdown/pull/689/commits/a3fbfa1edc04e3edfa981c5f61fde466a75bbae9)

apple/MarkdownTextInputDecoratorComponentView.mm
j-piasecki: Ah, good point - I added reseting it in `prepareForRecycle`

tomekzaw:
Unfortunately this PR causes a short delay when toggling `multiline` prop.

https://github.com/user-attachments/assets/4a5405bf-5f17-42a1-892a-2673d5b55b35

Also, toggling link color doesn't work after toggling `multiline` prop.

j-piasecki:
Should be fixed, though I needed to revert the asserts as the attach/detach methods can be called twice now: when the view is mounted for the very first time and when it's destroyed.

https://github.com/user-attachments/assets/4683a16c-6721-4cf6-8f9c-2c43cfabe5d2

apple/MarkdownTextInputDecoratorComponentView.mm
```diff
@@ -173,6 +203,12 @@ - (void)applyNewStyles
...
 }
 
+- (void)prepareForRecycle
+{
+  [super prepareForRecycle];
+  _markdownUtils = [[RCTMarkdownUtils alloc] init];
```
tomekzaw: If I understand correctly, the implementation of `prepareForRecycle` should be somewhat similar to `initWithFrame`. Do we need to reset `_props` and `_observersAdded` here as well?

apple/MarkdownTextInputDecoratorComponentView.mm
tomekzaw: Looks like the changes from a3fbfa1 are not applied now, could you please re-apply them?

j-piasecki: > though I needed to revert the asserts as the attach/detach methods can be called twice now: when the view is mounted for the very first time and when it's destroyed

tomekzaw: 😢 Okay

j-piasecki: Actually, I'm pretty sure we don't need to listen for window changes at all (we can keep observers attached on non-foreground screens), and listening for subview changes is enough: [3825919](https://github.com/Expensify/react-native-live-markdown/pull/689/commits/38259195031bba2c86d89b70e2e5547e0e30433c). In that case we can restore assertions.

apple/MarkdownTextInputDecoratorComponentView.mm
j-piasecki: I don't think we need to reset props since on the `Insert` mutation, `updateProps` will be called with the `oldProps` being a `nullptr`: https://github.com/facebook/react-native/blob/d631ec9a6c3cd2b5321de9ee35553878f1428b05/packages/react-native/React/Fabric/Mounting/RCTMountingManager.mm#L73-L91. I added it to be safe though.

When it comes to `_observersAdded` - no. It should never be true at this point, I've added an assert to reflect that.

tomekzaw:
I found this bug with singleline `MarkdownTextInput` but it is also reproducible on current main (5d999dfc).

https://github.com/user-attachments/assets/7bb1ecb9-85e3-4b6e-bd95-0c29b69a4869

PR #713, opened 2025-07-18
Bump React Native to 0.80.1

tomekzaw:
### Details
This PR bumps `react-native` to 0.80.1 according to https://react-native-community.github.io/upgrade-helper/?from=0.79.2&to=0.80.1 as well as upgrades `react-native-reanimated` to 3.18.0 for the compatibility with RN 0.80.

### Related Issues

GH_LINK


android/src/main/new_arch/react/renderer/components/RNLiveMarkdownSpec/MarkdownTextInputDecoratorShadowNode.cpp
```diff
@@ -3,7 +3,9 @@
 #include <fbjni/fbjni.h>
 #include <react/fabric/JFabricUIManager.h>
 #include <react/jni/ReadableNativeMap.h>
+#if REACT_NATIVE_MINOR_VERSION < 80
 #include <react/jni/SafeReleaseJniRef.h>
+#endif // REACT_NATIVE_MINOR_VERSION < 80
```
j-piasecki, line 8: ~~Why include it only on RN <0.80?~~

Nvm, I see

PR #715, opened 2025-07-23
Add support for React Native 0.81

j-piasecki:
### Details
Adapts the custom shadow node measurement logic and build setup to work on React Native 0.81. This PR should preserve compatibility with older versions of React Native, but **I haven't tested that**.

Also bumps dependencies:
- `react-native` to 0.81.0-rc.2
- `react-native-reanimated` to 3.19.0

### Related Issues

GH_LINK

### Manual Tests
Check the example app in general.

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -31,7 +31,7 @@ jobs:
       cancel-in-progress: true
     steps:
       - name: Select Xcode version
-        run: sudo xcode-select -s /Applications/Xcode_16.1.app
+        run: sudo xcode-select -s /Applications/Xcode_16.4.app
 
       - name: Check out Git repository
         # v4


.github/workflows/build-ios.yml
```diff
@@ -25,11 +25,14 @@ on:
...
     if: github.repository == 'Expensify/react-native-live-markdown'
-    runs-on: macos-13
+    runs-on: macos-15
     concurrency:
       group: build-ios-${{ github.ref }}
       cancel-in-progress: true
     steps:
+      - uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: 16.1
```
tomekzaw: Instead of relying on third-party actions, you can just use `xcode-select` command directly.

https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md#xcode

```suggestion
      - run: sudo xcode-select -s /Applications/Xcode_16.1.app
```

android/src/main/new_arch/CMakeLists.txt
```diff
@@ -43,16 +43,26 @@ target_link_libraries(
...
+if (ReactAndroid_VERSION_MINOR LESS_EQUAL 80)
+        target_compile_options(
+                ${LIB_TARGET_NAME}
+                PRIVATE
+                -DLOG_TAG=\"ReactNative\"
+                -fexceptions
+                -frtti
+                -Wall
+                -std=c++20
+                -DREACT_NATIVE_MINOR_VERSION=${ReactAndroid_VERSION_MINOR}
```
tomekzaw: Let's move `target_compile_options(... -DREACT_NATIVE_MINOR_VERSION=...)` outside of the conditional.

package.json
```diff
@@ -119,7 +120,7 @@
     "./WebExample"
   ],
   "engines": {
-    "node": ">= 18.0.0"
+    "node": ">= 22.14.0"
```
war-in: Do we have to bump the required node version? 🤔

j-piasecki: We probably shouldn't if we want compatibility with older RN versions. Restored to 18.0.0.

PR #720, opened 2025-08-13
[iOS & Android] Enable borderRadius in all mention types

war-in:
### Details

This PR adds support for `borderRadius` prop in mentions styles on iOS & Android

image

### Related Issues

https://github.com/Expensify/App/issues/19299

### Manual Tests

iOS & Android:

1. Verify if singleline mentions have rounded edges
2. Verify if multiline mentions have rounded edges only at the beginning and the end
3. Verify mentions work correctly in blockquotes

### DIFF
--- a/apple/MarkdownTextInputDecoratorComponentView.mm
+++ b/apple/MarkdownTextInputDecoratorComponentView.mm
@@ -101,14 +101,16 @@ using namespace facebook::react;
     [_markdownTextFieldObserver textFieldDidChange:_textField];
 
     if (@available(iOS 16.0, *)) {
-      std::string text = "txet";
-      std::reverse(text.begin(), text.end());
-      NSTextStorage *textStorage = [_textField valueForKey:[NSString stringWithFormat:@"%sStorage", text.c_str()]];
-      NSTextContainer *textContainer = [_textField valueForKey:[NSString stringWithFormat:@"%sContainer", text.c_str()]];
-      NSTextLayoutManager *textLayoutManager = [textContainer valueForKey:[NSString stringWithFormat:@"%sLayoutManager", text.c_str()]];
-
+      auto key = [](std::string s) {
+        std::reverse(s.begin(), s.end());
+        return @(s.c_str());
+      };
+      
+      NSTextContainer *textContainer = [_textField valueForKey:key("reniatnoCtxet_")];
+      NSTextLayoutManager *textLayoutManager = [textContainer valueForKey:key("reganaMtuoyaLtxet_")];
+      
       _markdownTextLayoutManagerDelegate = [[MarkdownTextLayoutManagerDelegate alloc] init];
-      _markdownTextLayoutManagerDelegate.textStorage = textStorage;
+      _markdownTextLayoutManagerDelegate.textStorage = [_textField valueForKey:key("egarotStxet_")];
       _markdownTextLayoutManagerDelegate.markdownUtils = _markdownUtils;
       textLayoutManager.delegate = _markdownTextLayoutManagerDelegate;
     }


src/styleUtils.ts
```diff
@@ -62,14 +62,17 @@ function makeDefaultMarkdownStyle(): MarkdownStyle {
     mentionHere: {
       color: 'green',
       backgroundColor: 'lime',
+      borderRadius: 5.0,
```
tomekzaw: Let's skip `.0` everywhere.
```suggestion
      borderRadius: 5,
```

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+      
+      CGPoint startLocation = [lineFragment locationForCharacterAtIndex:intersection.location];
+      CGPoint endLocation = [lineFragment locationForCharacterAtIndex:intersection.location + intersection.length];
+      
+      CGRect paddedRect = CGRectMake(startLocation.x,
+                                     lineBounds.origin.y,
+                                     endLocation.x - startLocation.x,
+                                     lineBounds.size.height);
+      
+      UIRectCorner cornersToRound = 0;
```
tomekzaw, line 112: Perhaps we should add an optimization that skips applying the rounded corners if border radius is set to `0` or `undefined`. We can either add some conditionals here or just set `NSBackgroundColorAttributeName` attribute. I'm wondering which approach is more suitable and I think the first approach makes more sense – this way we'll use `NSBackgroundColorAttributeName` only on iOS 15 and we'll still use `MarkdownTextLayoutFragment` mechanism no matter if the border radius is positive or zero. What do you think?

apple/RCTMarkdownStyle.mm
tomekzaw, line 1: Would it be possible to draw the rectangle with rounded corners below the selection layer so that the mention doesn't cover the selection? I know this is not possible using `NSBackgroundColorAttributeName`. As an alternative, we can always use a semi-transparent darker color instead of solid light color so the selection is at least partially visible. For the record, this is how it looks now:
 
obraz

src/styleUtils.ts
tomekzaw, line 1: Do we want to postpone merging this PR until the Android part is implemented?

apple/MarkdownFormatter.mm
```diff
@@ -91,14 +91,35 @@ - (void)applyRangeToAttributedString:(NSMutableAttributedString *)attributedStri
     [attributedString addAttribute:NSBackgroundColorAttributeName value:markdownStyle.codeBackgroundColor range:range];
   } else if (type == "mention-here") {
     [attributedString addAttribute:NSForegroundColorAttributeName value:markdownStyle.mentionHereColor range:range];
-    [attributedString addAttribute:NSBackgroundColorAttributeName value:markdownStyle.mentionHereBackgroundColor range:range];
+    if (@available(iOS 16.0, *)) {
+      [attributedString addAttribute:RCTLiveMarkdownMentionAttributeName
+                               value:@{@"backgroundColor": markdownStyle.mentionHereBackgroundColor,
+                                       @"cornerRadius": @(markdownStyle.mentionHereBorderRadius)}
```
tomekzaw: This is a nit-pick and I don't think it's necessary but how about using a specific struct like `RCTMarkdownMention` with non-null fields instead of a generic `NSDictionary`?

apple/MarkdownTextLayoutFragment.h
```diff
@@ -4,11 +4,10 @@
 NS_ASSUME_NONNULL_BEGIN
 
 API_AVAILABLE(ios(15.0))
-@interface BlockquoteTextLayoutFragment : NSTextLayoutFragment
+@interface MarkdownTextLayoutFragment : NSTextLayoutFragment
 
 @property (nonnull, atomic) RCTMarkdownUtils *markdownUtils;
-
-@property NSUInteger depth;
+@property (nonnull, atomic) NSNumber* depth;
```
tomekzaw: What's the reason behind this change?

Also, let's make it `NSNumber *depth` instead of `NSNumber* depth`.

apple/MarkdownTextLayoutManagerDelegate.mm
```diff
@@ -9,14 +9,34 @@ - (NSTextLayoutFragment *)textLayoutManager:(NSTextLayoutManager *)textLayoutMan
...
     NSNumber *depth = [self.textStorage attribute:RCTLiveMarkdownBlockquoteDepthAttributeName atIndex:index effectiveRange:nil];
-    if (depth != nil) {
-      BlockquoteTextLayoutFragment *textLayoutFragment = [[BlockquoteTextLayoutFragment alloc] initWithTextElement:textElement range:textElement.elementRange];
+    BOOL hasMention = [self hasMention:textElement];
+    
+    if (depth != nil || hasMention) {
+      MarkdownTextLayoutFragment *textLayoutFragment = [[MarkdownTextLayoutFragment alloc] initWithTextElement:textElement range:textElement.elementRange];
       textLayoutFragment.markdownUtils = _markdownUtils;
-      textLayoutFragment.depth = [depth unsignedIntValue];
+      textLayoutFragment.depth = depth;
```
tomekzaw: What's the reason behind this change? We can just pass `0` instead of `nil` and skip the blockquote rendering logic.

apple/MarkdownTextLayoutManagerDelegate.mm
```diff
@@ -9,14 +9,34 @@ - (NSTextLayoutFragment *)textLayoutManager:(NSTextLayoutManager *)textLayoutMan
...
       return textLayoutFragment;
     }
   }
   return [[NSTextLayoutFragment alloc] initWithTextElement:textElement range:textElement.elementRange];
 }
 
+- (BOOL)hasMention:(NSTextElement *)textElement {
+  NSTextParagraph *paragraph = (NSTextParagraph *)textElement;
```
tomekzaw: Which type would be more suitable for the method signature here, `NSTextElement *` or `NSTextParagraph *`?

apple/MarkdownTextLayoutManagerDelegate.mm
```diff
@@ -9,14 +9,34 @@ - (NSTextLayoutFragment *)textLayoutManager:(NSTextLayoutManager *)textLayoutMan
...
+  NSAttributedString *attributedString = [paragraph attributedString];
+  
+  __block BOOL hasMention = NO;
```
tomekzaw: I don't really like it that we use `hasMention` as the name of the method, local variable and the variable that holds the result. How about renaming the method to `containsMentionAttribute:` and renaming the local variable to `result`? Also, why don't we make the method accept just a `NSAttributedString *` so the logic is easier and move the text fragment conversions to the caller side?

apple/MarkdownFormatter.h
```diff
@@ -8,6 +8,8 @@ const NSAttributedStringKey RCTLiveMarkdownTextAttributeName = @"RCTLiveMarkdown
 
 const NSAttributedStringKey RCTLiveMarkdownBlockquoteDepthAttributeName = @"RCTLiveMarkdownBlockquoteDepth";
 
+const NSAttributedStringKey RCTLiveMarkdownMentionAttributeName = @"RCTLiveMarkdownMention";
```
tomekzaw: I think it's okay to keep the current name (which represents the type of syntax rather than the actual visual style) but ultimately I'd like to have it renamed to something which is not tied to the syntax elements once the library supports custom types of styles (for instance, let's say the developer will define `price` style which is rendered as background color with rounded rectangles, just like in CSS).

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+  return CGRectUnion(self.boundingRect, [super renderingSurfaceBounds]);
+}
+
+- (void)drawAtPoint:(CGPoint)point inContext:(CGContextRef)ctx {
+  if (self.textLineFragments.count == 0) {
+    [super drawAtPoint:point inContext:ctx];
+    return;
+  }
+  
+  [self drawRibbon];
```
tomekzaw:
```suggestion
  [self drawBlockquotes];
```
or
```suggestion
  [self drawBlockquoteRibbons];
```

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+  CGFloat marginLeft = _markdownUtils.markdownStyle.blockquoteMarginLeft;
+  CGFloat borderWidth = _markdownUtils.markdownStyle.blockquoteBorderWidth;
+  CGFloat paddingLeft = _markdownUtils.markdownStyle.blockquotePaddingLeft;
+  CGFloat shift = marginLeft + borderWidth + paddingLeft;
+  
+  [_markdownUtils.markdownStyle.blockquoteBorderColor setFill];
+  
+  CGRect boundingRect = self.boundingRect;
+  for (NSUInteger i = 0; i < [_depth unsignedIntValue]; ++i) {
+    CGRect ribbonRect = CGRectMake(boundingRect.origin.x + i * shift, boundingRect.origin.y, borderWidth, boundingRect.size.height);
```
tomekzaw: I know this code has been copied but can we move `boundingRect.origin.x + i * shift` calculation to a separate line of code, like:
```suggestion
    CGFloat x = boundingRect.origin.x + i * shift;
    CGRect ribbonRect = CGRectMake(x, boundingRect.origin.y, borderWidth, boundingRect.size.height);
```
Would be good to also rename `i` to `level`.

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+    UIRectFill(ribbonRect);
+  }
+}
+
+- (void)drawMentions {
+  NSMutableArray<NSDictionary *> *mentions = [self getMentions];
```
tomekzaw: It seems like we only use `mentions` once, how about moving `[self getMentions]` to the place where it is used?

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+  [self.textLineFragments enumerateObjectsUsingBlock:^(NSTextLineFragment * _Nonnull lineFragment, NSUInteger idx, BOOL * _Nonnull stop) {
+    if (lineFragment.characterRange.length == 0) {
+      return;
+    }
+    
+    CGRect lineBounds = lineFragment.typographicBounds;
+    for (NSDictionary *mention in mentions) {
+      NSRange mentionRange = [mention[@"range"] rangeValue];
+      UIColor *backgroundColor = mention[@"value"][@"backgroundColor"];
+      CGFloat cornerRadius = [mention[@"value"][@"cornerRadius"] floatValue];
```
tomekzaw:
```suggestion
    for (RCTMarkdownMention *mention in mentions) {
      NSRange mentionRange = mention.range;
      UIColor *backgroundColor = mention.backgroundColor;
      CGFloat cornerRadius = mention.cornerRadius;
```

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+      
+      NSRange intersection = NSIntersectionRange(lineFragment.characterRange, mentionRange);
```
tomekzaw: Would be good to use `UITextRange *` instead of `NSRange` if possible

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+      
+      BOOL isStart = (intersection.location == mentionRange.location);
+      BOOL isEnd = (NSMaxRange(intersection) == NSMaxRange(mentionRange));
+      
+      CGPoint startLocation = [lineFragment locationForCharacterAtIndex:intersection.location];
+      CGPoint endLocation = [lineFragment locationForCharacterAtIndex:intersection.location + intersection.length];
+      
+      CGRect paddedRect = CGRectMake(startLocation.x,
+                                     lineBounds.origin.y,
+                                     endLocation.x - startLocation.x,
```
tomekzaw: How about moving `endLocation.x - startLocation.x` to a separate variable named `width`?

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+                                     lineBounds.size.height);
+      
+      UIRectCorner cornersToRound = 0;
+      if (isStart && isEnd) {
+        cornersToRound = UIRectCornerAllCorners;
+      } else if (isStart) {
+        cornersToRound = (UIRectCornerTopLeft | UIRectCornerBottomLeft);
+      } else if (isEnd) {
+        cornersToRound = (UIRectCornerTopRight | UIRectCornerBottomRight);
+      }
```
tomekzaw: It seems like `UIRectCorner(...)` are bit masks so most likely `UIRectCornerAllCorners` is numerically equal to `UIRectCornerTopLeft | UIRectCornerBottomLeft | UIRectCornerTopRight | UIRectCornerBottomRight`. If that's the case indeed, we can just do
```suggestion
      if (isStart) {
        cornersToRound |= UIRectCornerTopLeft | UIRectCornerBottomLeft;
      }
      if (isEnd) {
        cornersToRound |= UIRectCornerTopRight | UIRectCornerBottomRight;
      }
```

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+      
+      UIBezierPath *linePath;
+      linePath = [UIBezierPath bezierPathWithRoundedRect:paddedRect
+                                       byRoundingCorners:cornersToRound
+                                             cornerRadii:CGSizeMake(cornerRadius, cornerRadius)];
```
tomekzaw:
```suggestion
      UIBezierPath *linePath = [UIBezierPath bezierPathWithRoundedRect:paddedRect
                                       byRoundingCorners:cornersToRound
                                             cornerRadii:CGSizeMake(cornerRadius, cornerRadius)];
```

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+
+- (CGRect)boundingRect {
+  CGRect fragmentTextBounds = CGRectNull;
+  for (NSTextLineFragment *lineFragment in self.textLineFragments) {
+    if (lineFragment.characterRange.length == 0) {
+      continue;
+    }
+    CGRect lineFragmentBounds = lineFragment.typographicBounds;
+    if (CGRectIsNull(fragmentTextBounds)) {
+      fragmentTextBounds = lineFragmentBounds;
```
tomekzaw, line 143: Do we really want to overwrite `fragmentTextBounds` (outside of for-loop) with the last `lineFragmentBounds`?

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+  
+  CGFloat marginLeft = _markdownUtils.markdownStyle.blockquoteMarginLeft;
+  CGFloat borderWidth = _markdownUtils.markdownStyle.blockquoteBorderWidth;
+  CGFloat paddingLeft = _markdownUtils.markdownStyle.blockquotePaddingLeft;
+  CGFloat shift = marginLeft + borderWidth + paddingLeft;
+  
+  fragmentTextBounds.origin.x -= (paddingLeft + borderWidth) + shift * ([_depth unsignedIntValue] - 1);
+  fragmentTextBounds.size.width = borderWidth + shift * ([_depth unsignedIntValue] - 1);
```
tomekzaw: Looks like the blockquote logic is now duplicated, can we somehow dedupe it?

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+  
+  return fragmentTextBounds;
+}
+
+- (NSMutableArray<NSDictionary *>*)getMentions {
```
tomekzaw: It seems a bit suspicious that this method doesn't accept any arguments. Also, it's called only once, so why don't we just inline it in the place where it is used for better code readability?

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,145 @@
...
+  NSMutableArray<NSDictionary *> *mentions = [NSMutableArray array];
+  [attributedString enumerateAttribute:RCTLiveMarkdownMentionAttributeName
+                               inRange:NSMakeRange(0, attributedString.length)
+                               options:0
+                            usingBlock:^(id value, NSRange range, BOOL *stop) {
+    if (value) {
+      [mentions addObject:@{
+        @"range": [NSValue valueWithRange:range],
+        @"value": value
+      }];
```
tomekzaw: Let's use a struct instead

apple/MarkdownTextLayoutFragment.mm
war-in: This is the code from the blockquotes file so I'm not sure
I just moved it here but can take a look if you want

apple/MarkdownTextLayoutFragment.mm
war-in: We use it a few times if the paragraph has more than one line

```
[self.textLineFragments enumerateObjectsUsingBlock:^(NSTextLineFragment * _Nonnull lineFragment, NSUInteger idx, BOOL * _Nonnull stop) {
```

apple/MarkdownTextLayoutFragment.mm
war-in: Unfortunately, both `lineFragment.characterRange` and `mentionRange` are `NSRange`s so we would have to convert them, which I think is not worth the struggle

apple/MarkdownTextLayoutFragment.mm
war-in: It was already :/
I could create a helper function to get the shift to make it look better but the logic would stay unchanged

I think we need it to be that way because boundingRect is also used in renderingSurfaceBounds

apple/MarkdownFormatter.h
```diff
@@ -6,6 +6,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 const NSAttributedStringKey RCTLiveMarkdownTextAttributeName = @"RCTLiveMarkdownText";
 
+const NSAttributedStringKey RCTLiveMarkdownTextBackgroundAttributeName = @"RCTLiveMarkdownTextBackground";
+@interface RCTMarkdownTextBackground : NSObject
+  @property (nonatomic, strong) UIColor *color;
+  @property (nonatomic, assign) CGFloat cornerRadius;
+@end
```
war-in: I'm not sure if this is the right place to put this object in
@tomekzaw do you have a better idea? Maybe a separate file for RCTMarkdownTextBackground and RCTMarkdownTextBackgroundWithRange? 🤔

src/styleUtils.ts
war-in: yes, I think we should keep both platforms aligned

apple/RCTMarkdownStyle.mm
war-in: I think the alternative is the way to go as I couldn't find the right solution (to draw it under the selection layer)

apple/MarkdownFormatter.h
tomekzaw: We can add that in a separate file

android/src/main/java/com/expensify/livemarkdown/spans/MarkdownBackgroundSpan.java
```diff
@@ -0,0 +1,87 @@
...
+
+import androidx.annotation.ColorInt;
+import androidx.annotation.NonNull;
+
+public class MarkdownBackgroundSpan implements MarkdownSpan, LineBackgroundSpan {
+
+  private final int backgroundColor;
+  private final int mentionStart;
+  private final int mentionEnd;
+  private final float cornerRadius;
```
tomekzaw:
```suggestion
  private final float borderRadius;
```

war-in:
I fixed last known iOS issue in this commit - https://github.com/Expensify/react-native-live-markdown/pull/720/commits/28b6a13e75c87782d3fc54824d47329db9599fae

Before
image

After
image

src/styleUtils.ts
```diff
@@ -62,14 +62,17 @@ function makeDefaultMarkdownStyle(): MarkdownStyle {
     mentionHere: {
       color: 'green',
       backgroundColor: 'lime',
+      borderRadius: 5,
```
jmusial, line 65: Is it me or default `borderRadius` for android and web look different ? 

image

jmusial (commented):
Dunno if those are connected to this PR, or just existing bugs. In all cases, it's just on mobile devices 

1. android, switching to single line breaks highlights :( 

https://github.com/user-attachments/assets/02278d06-da9d-486c-b7e3-3dfcd6a21dde

2. android, Cursor not visible when within `codeblock` or `mention` 

https://github.com/user-attachments/assets/e8106b62-b606-4005-9a9d-b9d7a45b996d

jmusial:
Very much a nitpick, but when having a multiline mention starting bottom border radius and finish top, should not apply. Same issue for web 

image

jmusial:
Na ios dzieją się rzeczy niestworzone :( Ale to też single - line

to jest `17.5`, nie wiem czy to ma znaczenie

https://github.com/user-attachments/assets/2fc0da71-3841-4968-82b3-cbb8e1a62b8d

war-in:
> Very much a nitpick, but when having a multiline mention starting bottom border radius and finish top, should not apply. Same issue for web

We can try to fix it in a separate PR but I think we need to consult with the design team. I'll leave it as an open idea because it's not in the scope of this PR

Thanks for noticing!

src/styleUtils.ts
war-in: Yeah, saw that too. Not sure why the same radius is smaller on Android 🤔 

@tomekzaw any ideas?

war-in:
> 1. android, switching to single line breaks highlights :(

@jmusial should be fixed with the latest commit 🎉 https://github.com/Expensify/react-native-live-markdown/pull/720/commits/98dafdaaf9f2d9bd92df4dcd3e0c88222d481469

> 2. android, Cursor not visible when within codeblock or mention

this is expected behaviour. We discussed it with the design team and there is no plan to support it in the near future

tomekzaw:
> 2. android, Cursor not visible when within codeblock or mention

For the context, this was discussed on Slack: https://swmansion.slack.com/archives/C01GTK53T8Q/p1755699466781499

apple/MarkdownTextInputDecoratorComponentView.mm
```diff
@@ -100,6 +100,17 @@ - (void)addTextInputObservers
     // format initial value
     [_markdownTextFieldObserver textFieldDidChange:_textField];
 
+    if (@available(iOS 16.0, *)) {
+      NSTextStorage *textStorage = [_textField valueForKey:@"_textStorage"];
+      NSTextContainer *textContainer = [_textField valueForKey:@"_textContainer"];
+      NSTextLayoutManager *textLayoutManager = [textContainer valueForKey:@"_textLayoutManager"];
```
war-in: I'm not sure if accessing UITextField's private fields could cause some rejections on app store. We should probably verify that

cc @tomekzaw any ideas on how to do that?

src/styleUtils.ts
war-in: fixed here - https://github.com/Expensify/react-native-live-markdown/pull/720/commits/39e031876c225ca6230bbc9684dee8fb43f5a293

war-in:
@tomekzaw @parasharrajat @Skalakid PR's ready for review 🎉 

Please review it carefully and ask me any questions because the solution is quite complicated in a few places. If you think adding comments would be valuable let me know!

android/src/main/java/com/expensify/livemarkdown/spans/MarkdownBackgroundSpan.java
```diff
@@ -0,0 +1,94 @@
...
+    CharSequence lineText = text.subSequence(start, end);
+    if (layout == null || layout.getText() != lineText || layout.getWidth() != right || layout.getLineEnd(0) != lineEnd) {
+      // Create layout for the current line only
+      layout = StaticLayout.Builder.obtain(lineText, lineStart, lineEnd, (TextPaint) paint, right).build();
+
+      int relativeMentionStart = mentionStart - start;
+      int relativeMentionEnd = mentionEnd - start;
+
+      boolean mentionStarts = lineStart <= relativeMentionStart;
+      boolean mentionEnds = lineEnd >= relativeMentionEnd;
```
Skalakid: How about renaming these variables to something for example like: `startsInThisLine` and `endsInThisLine`

apple/MarkdownFormatter.mm
```diff
@@ -91,14 +92,44 @@ - (void)applyRangeToAttributedString:(NSMutableAttributedString *)attributedStri
...
+      RCTMarkdownTextBackground *textBackground = [[RCTMarkdownTextBackground alloc] init];
+      textBackground.color = markdownStyle.mentionHereBackgroundColor;
+      textBackground.borderRadius = markdownStyle.mentionHereBorderRadius;
+      
+      [attributedString addAttribute:RCTLiveMarkdownTextBackgroundAttributeName
+                               value:textBackground
+                               range:range];
+    } else {
+      [attributedString addAttribute:NSBackgroundColorAttributeName value:markdownStyle.mentionHereBackgroundColor range:range];
+    }
```
Skalakid: I see that there is a lot of similar code. How about moving it to a separate function with textColor, backgroundColor, and borderRadius as props (etc)

apple/MarkdownTextLayoutFragment.mm
```diff
@@ -0,0 +1,159 @@
...
+      NSRange mentionRange = mention.range;
+      
+      if (mentionRange.location >= NSMaxRange(lineRange)) {
+          break;
+      }
+      
+      NSRange intersection = NSIntersectionRange(lineRange, mentionRange);
+      CGPoint startLocation = [lineFragment locationForCharacterAtIndex:intersection.location];
+      if (isSingleline && startLocation.x == 0 && intersection.location > 0) {
+        // singleline: mention starts off screen, no need to draw background
```
Skalakid, line 81: Please add an explanation that it happens when the input isn't focused

android/src/main/java/com/expensify/livemarkdown/spans/MarkdownBackgroundSpan.java
Skalakid, line 1: The amount of different start and end variables is a bit overwhelming there 😓. Maybe let's add some general comments or something?

android/src/main/java/com/expensify/livemarkdown/MarkdownStyle.java
```diff
@@ -95,10 +103,13 @@ public MarkdownStyle(@NonNull ReadableMap map, @NonNull Context context) {
     mPreBackgroundColor = parseColor(map, "pre", "backgroundColor", context);
     mMentionHereColor = parseColor(map, "mentionHere", "color", context);
     mMentionHereBackgroundColor = parseColor(map, "mentionHere", "backgroundColor", context);
+    mMentionHereBorderRadius = parseFloat(map, "mentionHere", "borderRadius") * screenDensity;
```
tomekzaw, line 106: I'd rather keep the original value of `borderRadius` in `MarkdownStyle` and multiply it by `screenDensity` near the drawing logic.

apple/MarkdownTextInputDecoratorComponentView.mm
tomekzaw: We need to obfuscate all the selectors here, e.g. `"_text" + "tuoyal".reversed() + Manager"`

apple/MarkdownFormatter.mm
tomekzaw: Helper function for applying mention-related attributes sounds good!

apple/MarkdownTextLayoutFragment.h
```diff
@@ -1,15 +1,18 @@
...
 
 NS_ASSUME_NONNULL_BEGIN
 
 API_AVAILABLE(ios(15.0))
-@interface BlockquoteTextLayoutFragment : NSTextLayoutFragment
+@interface MarkdownTextLayoutFragment : NSTextLayoutFragment
 
 @property (nonnull, atomic) RCTMarkdownUtils *markdownUtils;
 
 @property NSUInteger depth;
```
tomekzaw, line 12: At some point we might wanna rename `depth` to `blockquoteDepth` as this is unrelated to `mentions`. This can be done in a separate PR (to be merged before or after this one).

apple/RCTMarkdownTextBackgroundWithRange.h
tomekzaw, line 1: I'd suggest moving these two classes to separate files, i.e. `RCTMarkdownTextBackground.h` and `RCTMarkdownTextBackground.mm` (and the same for `RCTMarkdownTextBackgroundWithRange`).

android/src/main/java/com/expensify/livemarkdown/MarkdownStyle.java
war-in: I tried that but context is not available in span so I had to pass it down and the code looked messy
We can do that but this approach is much cleaner

war-in:
@parasharrajat kind bump :)

src/styleUtils.ts
parasharrajat, line 65: I think it should be 4px as per design system.

WebExample/__tests__/styles.spec.ts
```diff
@@ -40,15 +40,15 @@ test.describe('markdown content styling', () => {
   });
 
   test('mention-here', async ({page}) => {
-    await testMarkdownContentStyle({testContent: 'here', style: 'color: green; background-color: lime;', page});
+    await testMarkdownContentStyle({testContent: 'here', style: 'color: green; background-color: lime; border-radius: 5px;', page});
```
parasharrajat, line 43: Need to be 4px here too.

src/styleUtils.ts
tomekzaw: We're using different values here on purpose. For instance, we also have background color `lime` and foreground color `green`. That's because they need to be set on the E/App side.

parasharrajat: Make sense. I didn't notice that it is a different repo. 😄

war-in: @parasharrajat so are we ready for the ✅? 👀

WebExample/__tests__/styles.spec.ts
war-in: discussed in the previous [comment](https://github.com/Expensify/react-native-live-markdown/pull/720#discussion_r2554232012)

war-in:
@parasharrajat can we get a final ✅? All comments have been answered!

parasharrajat:
BUG: Android: Cursor shows behind the background. 

https://github.com/user-attachments/assets/8a8b9713-1d3f-4925-8c14-3424ef687a82

parasharrajat:
How do I build iOS build locally? 

```
"@expensify/react-native-live-markdown": "git://github.com/Expensify/react-native-live-markdown#eb521b53582165d532918ca64b859210dc92bf09",
```

I installed the latest commit for this package on NewDot repo, and it worked for Android. But for iOS, the app is downloading the app from the remote cache.

war-in:
> [BUG: Android: Cursor shows behind the background](https://github.com/Expensify/react-native-live-markdown/pull/720#issuecomment-3602206125)

We know about the issue and already [reported it](https://swmansion.slack.com/archives/C01GTK53T8Q/p1755699466781499). It was put on hold at the time but we could revisit this idea in another issue/thread.
Marking as unrelated for now!

war-in:
> But for iOS, the app is downloading the app from the remote cache.

@parasharrajat I'd probably remove the GITHUB_TOKEN entry from my `.env` to achieve that. This should be the easiest way to disable remote builds

parasharrajat:
~Nice idea. I will try that.~ Won't work


Now the builds are coming from S3.

war-in:
Ok, so another quick one - I'd add a dummy log to some ~android~ iOS file to change the fingerprint. That should work

parasharrajat:
iOS looks good.

parasharrajat:
@war-in So we are not targeting this https://github.com/Expensify/react-native-live-markdown/pull/720#issuecomment-3602206125 here in this PR. Do we have approval for this from internal team?

@puneetlath What do you say?

parasharrajat:
### Screenshots

#### :black_square_button: iOS / native

https://github.com/user-attachments/assets/2069491a-8ce7-493d-a1c8-5c1c528fbd0f

#### :black_square_button: MacOS / Chrome
image


#### :black_square_button: Android / native

https://github.com/user-attachments/assets/e4949c67-55f7-4a58-ba87-93ecd2891d67

war-in:
> @war-in So we are not targeting this https://github.com/Expensify/react-native-live-markdown/pull/720#issuecomment-3602206125 here in this PR. Do we have approval for this from internal team?

> Do we have approval for this

@parasharrajat which `this` are you referring to here? approval for not including the fix in this PR or approval for handling this in another issue?

parasharrajat:
> approval for not including the fix in this PR or approval for handling this in another issue?

Both are the same. We will have to fix that. The question is where. This Pr or another issue. 

So we need to clarify that first before moving forward.

puneetlath:
Sorry, the Slack link above doesn't work for me. Is that a bug this PR introduces? Or it already exists?

war-in:
> introduced by this.

@parasharrajat No, it's not. I downloaded the prod version from google play and this is how it behaves right now (see not-rounded corners)
So this behaviour is not introduced by this PR

https://github.com/user-attachments/assets/7e9e80b7-c2f4-4b7f-84c9-eea2f8173c30

cc @puneetlath (you can enter the [slack link](https://swmansion.slack.com/archives/C01GTK53T8Q/p1755699466781499) by copying it and pasting in a slack conversation e.g. self dm)

parasharrajat:
Ah, I see. Thanks for correcting me. Then we are fine with this PR.

parasharrajat (approved):
:ribbon: :eyes: :ribbon:  C+ reviewed

PR #729, opened 2025-09-18
Fix moving cursor to the end of line

Skalakid:
### Details

This PR fixes moving the cursor to the end of the line using CMD + ➡ or the End key. To fix this problem, we implemented splitting italic and strikethrough markdown types on the starting and ending syntax of the inline code. Because of the HTML structure, we couldn't split the range directly on the inline code. So similarly to other applications like Slack, the styles are applied to the text inside inline code separately

### Related Issues

https://github.com/Expensify/App/issues/69493#issuecomment-3308063242

### DIFF
--- a/src/rangeUtils.ts
+++ b/src/rangeUtils.ts
@@ -58,6 +58,24 @@ function ungroupRanges(ranges: MarkdownRange[]): MarkdownRange[] {
 }
 
 /**
+ * Creates a list of ranges that should not be formatted by certain markdown types (italic, strikethrough).
+ * This includes emojis and syntaxes of inline code blocks.
+ */
+function getRangesToExcludeFormatting(ranges: MarkdownRange[]) {
+  let closingSyntaxPosition: number | null = null;
+  return ranges.filter((range, index) => {
+    const nextRange = ranges[index + 1];
+    if (nextRange && nextRange.type === 'code' && range.type === 'syntax') {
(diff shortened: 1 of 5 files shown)


src/rangeUtils.ts
```diff
@@ -57,8 +57,33 @@ function ungroupRanges(ranges: MarkdownRange[]): MarkdownRange[] {
...
+  });
+}
+
+/**
+ * Splits ranges of a specific type from being formatted by specified markdown types (e.g., 'emoji', 'syntax').
+ * @param ranges - The array of MarkdownRange objects to process.
+ * @param baseMarkdownType - The base markdown type to exclude formatting from (e.g., 'italic').
+ * @param rangesToExclude - The array of MarkdownRange objects representing the ranges to exclude from formatting.
+ */
+function excludeRangeTypesFromFormatting(ranges: MarkdownRange[], baseMarkdownType: MarkdownType, rangesToExclude: MarkdownRange[]): MarkdownRange[] {
```
jmusial, line 86: NAB  but `baseMarkdownType` could be an array of types maybe ?  would save one loop over `rangesToExclude` and one line in `parseExpensiMark` :)

Skalakid: It would require more changes in the current algorithm. Right now, we need to execute `excludeRangeTypesFromFormatting` one after another, because otherwise, split ranges start to overlap each other, and it breaks the markdown rendering then

Issue #726, opened 2025-09-04
Prepare live markdown for Reanimated v4 🚀


blazejkustra:
Here is the patch that I used locally, hope it helps @war-in!

```diff
diff --git a/node_modules/@expensify/react-native-live-markdown/RNLiveMarkdown.podspec b/node_modules/@expensify/react-native-live-markdown/RNLiveMarkdown.podspec
index 21de59d..da055d6 100644
--- a/node_modules/@expensify/react-native-live-markdown/RNLiveMarkdown.podspec
+++ b/node_modules/@expensify/react-native-live-markdown/RNLiveMarkdown.podspec
@@ -5,7 +5,7 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
 react_native_minor_version = react_native_json['version'].split('.')[1].to_i
 
 pods_root = Pod::Config.instance.project_pods_root
-react_native_reanimated_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] || File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('react-native-reanimated/package.json')"`)
+react_native_reanimated_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] || File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('react-native-worklets/package.json')"`)
 react_native_reanimated_node_modules_dir_from_pods_root = Pathname.new(react_native_reanimated_node_modules_dir).relative_path_from(pods_root).to_s
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
@@ -24,7 +24,7 @@ Pod::Spec.new do |s|
 
   s.source_files = "apple/**/*.{h,m,mm}", "cpp/**/*.{h,cpp}"
 
-  s.dependency "RNReanimated/worklets"
+  s.dependency "RNWorklets"
 
   s.xcconfig = {
     "OTHER_CFLAGS" => "$(inherited) -DREACT_NATIVE_MINOR_VERSION=#{react_native_minor_version}",
diff --git a/node_modules/@expensify/react-native-live-markdown/android/build.gradle b/node_modules/@expensify/react-native-live-markdown/android/build.gradle
index 001a23c..232263c 100644
--- a/node_modules/@expensify/react-native-live-markdown/android/build.gradle
[...]

PR #731, opened 2025-10-03
Migrate from `react-native-reanimated` to `react-native-worklets` with backward compatibility

war-in:
### Related Issues

GH_LINK https://github.com/Expensify/react-native-live-markdown/issues/726

### DIFF
--- a/RNLiveMarkdown.podspec
+++ b/RNLiveMarkdown.podspec
@@ -56,6 +56,9 @@ Pod::Spec.new do |s|
       "\"$(PODS_ROOT)/#{react_native_worklets_node_modules_dir_from_pods_root}/Common/cpp\"",
     ].join(' '),
   }
+  if is_worklets
+    s.xcconfig["OTHER_CFLAGS"] << " -DIS_WORKLETS=1"
+  end
 
   s.pod_target_xcconfig = { "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/cpp\"" }
 
--- a/RNLiveMarkdown.podspec
+++ b/RNLiveMarkdown.podspec
@@ -45,7 +45,7 @@ Pod::Spec.new do |s|
     s.dependency "RNReanimated/worklets"
   end
 
-  s.xcconfig = {
+  xcconfig = {
     "OTHER_CFLAGS" => "$(inherited) -DREACT_NATIVE_MINOR_VERSION=#{react_native_minor_version}",
     "HEADER_SEARCH_PATHS" => [
       "\"$(PODS_ROOT)/#{react_native_worklets_node_modules_dir_from_pods_root}/apple\"",
@@ -53,8 +53,9 @@ Pod::Spec.new do |s|
     ].join(' '),
   }
   if worklets_installed
-    s.xcconfig["OTHER_CFLAGS"] << " -DWORKLETS_INSTALLED=1"
+    xcconfig["OTHER_CFLAGS"] << " -DWORKLETS_INSTALLED=1"
   end
+  s.xcconfig = xcconfig
 
   s.pod_target_xcconfig = { "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/cpp\"" }
 
--- a/RNLiveMarkdown.podspec
+++ b/RNLiveMarkdown.podspec
@@ -5,8 +5,25 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
 react_native_minor_version = react_native_json['version'].split('.')[1].to_i
 
 pods_root = Pod::Config.instance.project_pods_root
-react_native_reanimated_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] || File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('react-native-reanimated/package.json')"`)
-react_native_reanimated_node_modules_dir_from_pods_root = Pathname.new(react_native_reanimated_node_modules_dir).relative_path_from(pods_root).to_s
+
+find_installed_package = ->(name) do
+  package_path = File.join(react_native_node_modules_dir, name)
+  File.directory?(package_path)
+end
+
+is_reanimated = find_installed_package.call('react-native-reanimated')
+is_worklets   = find_installed_package.call('react-native-worklets')
+
+package_name = if is_reanimated
+                 'react-native-reanimated/package.json'
+               elsif is_worklets
+                 'react-native-worklets/package.json'
+               else
+                 raise "Error!"
+               end
+
+react_native_worklets_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] || File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('#{package_name}')"`)
+react_native_worklets_node_modules_dir_from_pods_root = Pathname.new(react_native_worklets_node_modules_dir).relative_path_from(pods_root).to_s
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
@@ -24,13 +41,19 @@ Pod::Spec.new do |s|
 
   s.source_files = "apple/**/*.{h,m,mm}", "cpp/**/*.{h,cpp}"
 
-  s.dependency "RNReanimated/worklets"
+  if reanimated
+    s.dependency "RNReanimated/worklets"
+  elsif worklets
+    s.dependency "RNWorklets"
+  else
+    raise "Error!"
+  end
 
   s.xcconfig = {
     "OTHER_CFLAGS" => "$(inherited) -DREACT_NATIVE_MINOR_VERSION=#{react_native_minor_version}",
     "HEADER_SEARCH_PATHS" => [
-      "\"$(PODS_ROOT)/#{react_native_reanimated_node_modules_dir_from_pods_root}/apple\"",
-      "\"$(PODS_ROOT)/#{react_native_reanimated_node_modules_dir_from_pods_root}/Common/cpp\"",
+      "\"$(PODS_ROOT)/#{react_native_worklets_node_modules_dir_from_pods_root}/apple\"",
+      "\"$(PODS_ROOT)/#{react_native_worklets_node_modules_dir_from_pods_root}/Common/cpp\"",
     ].join(' '),
   }
 
--- a/RNLiveMarkdown.podspec
+++ b/RNLiveMarkdown.podspec
@@ -14,10 +14,10 @@ end
 is_reanimated = find_installed_package.call('react-native-reanimated')
 is_worklets   = find_installed_package.call('react-native-worklets')
 
-package_name = if is_reanimated
-                 'react-native-reanimated/package.json'
-               elsif is_worklets
+package_name = if is_worklets
                  'react-native-worklets/package.json'
+               elsif is_reanimated
+                 'react-native-reanimated/package.json'
                else
                  raise "Error!"
                end
@@ -41,10 +41,10 @@ Pod::Spec.new do |s|
 
   s.source_files = "apple/**/*.{h,m,mm}", "cpp/**/*.{h,cpp}"
 
-  if is_reanimated
-    s.dependency "RNReanimated/worklets"
-  elsif is_worklets
+  if is_worklets
     s.dependency "RNWorklets"
+  elsif is_reanimated
+    s.dependency "RNReanimated/worklets"
   else
     raise "Error!"
   end
--- a/RNLiveMarkdown.podspec
+++ b/RNLiveMarkdown.podspec
@@ -6,22 +6,20 @@ react_native_minor_version = react_native_json['version'].split('.')[1].to_i
 
 pods_root = Pod::Config.instance.project_pods_root
 
-find_installed_package = ->(name) do
+is_package_installed = ->(name) do
   package_path = File.join(react_native_node_modules_dir, name)
   File.directory?(package_path)
 end
 
-is_reanimated = find_installed_package.call('react-native-reanimated')
-is_worklets   = find_installed_package.call('react-native-worklets')
+worklets_installed = is_package_installed.call('react-native-worklets')
 
-package_name = if is_worklets
+package_name = if worklets_installed
                  'react-native-worklets/package.json'
-               elsif is_reanimated
-                 'react-native-reanimated/package.json'
                else
-                 raise "Error!"
+                 'react-native-reanimated/package.json'
                end
 
+# TODO: is REACT_NATIVE_REANIMATED_NODE_MODULES_DIR needed, should we rename it or add new variable?
 react_native_worklets_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] || File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('#{package_name}')"`)
 react_native_worklets_node_modules_dir_from_pods_root = Pathname.new(react_native_worklets_node_modules_dir).relative_path_from(pods_root).to_s
 
@@ -41,12 +39,10 @@ Pod::Spec.new do |s|
 
   s.source_files = "apple/**/*.{h,m,mm}", "cpp/**/*.{h,cpp}"
 
-  if is_worklets
+  if worklets_installed
     s.dependency "RNWorklets"
-  elsif is_reanimated
-    s.dependency "RNReanimated/worklets"
   else
-    raise "Error!"
+    s.dependency "RNReanimated/worklets"
   end
 
   s.xcconfig = {
@@ -56,8 +52,8 @@ Pod::Spec.new do |s|
       "\"$(PODS_ROOT)/#{react_native_worklets_node_modules_dir_from_pods_root}/Common/cpp\"",
     ].join(' '),
   }
-  if is_worklets
-    s.xcconfig["OTHER_CFLAGS"] << " -DIS_WORKLETS=1"
+  if worklets_installed
+    s.xcconfig["OTHER_CFLAGS"] << " -DWORKLETS_INSTALLED=1"
   end
 
   s.pod_target_xcconfig = { "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/cpp\"" }
--- a/RNLiveMarkdown.podspec
+++ b/RNLiveMarkdown.podspec
@@ -41,9 +41,9 @@ Pod::Spec.new do |s|
 
   s.source_files = "apple/**/*.{h,m,mm}", "cpp/**/*.{h,cpp}"
 
-  if reanimated
+  if is_reanimated
     s.dependency "RNReanimated/worklets"
-  elsif worklets
+  elsif is_worklets
     s.dependency "RNWorklets"
   else
     raise "Error!"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -107,7 +107,7 @@ android {
           "-DANDROID_TOOLCHAIN=clang",
           "-DREACT_NATIVE_MINOR_VERSION=${REACT_NATIVE_MINOR_VERSION}",
           "-DANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES=ON",
-          "-DREACT_NATIVE_ROOT_DIR=${resolveReactNativeDirectory()}"
+          "-DREACT_NATIVE_ROOT_DIR=${resolveReactNativeDirectory().path}"
         abiFilters (*reactNativeArchitectures())
       }
     }
@@ -190,10 +190,10 @@ dependencies {
   implementation "com.facebook.react:react-android" // version substituted by RNGP
   implementation "com.facebook.react:hermes-android" // version substituted by RNGP
 
-  if (isReanimated) {
-    implementation project(":react-native-reanimated")
-  } else if (isWorklets) {
+  if (isWorklets) {
     implementation project(":react-native-worklets")
+  } else if (isReanimated) {
+    implementation project(":react-native-reanimated")
   }
 }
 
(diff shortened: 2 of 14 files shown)


RNLiveMarkdown.podspec
```diff
@@ -5,8 +5,23 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
 react_native_minor_version = react_native_json['version'].split('.')[1].to_i
 
 pods_root = Pod::Config.instance.project_pods_root
-react_native_reanimated_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] || File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('react-native-reanimated/package.json')"`)
-react_native_reanimated_node_modules_dir_from_pods_root = Pathname.new(react_native_reanimated_node_modules_dir).relative_path_from(pods_root).to_s
+
+is_package_installed = ->(name) do
```
tomekzaw: Why not just ...?
```suggestion
def is_package_installed(name)
```

RNLiveMarkdown.podspec
```diff
@@ -5,8 +5,23 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
...
+  package_path = File.join(react_native_node_modules_dir, name)
+  File.directory?(package_path)
```
tomekzaw: `react-native-worklets` can technically be installed as a link and in that case `node_modules/react-native-worklets` will be a symlink instead of a directory. Let's find a better way to detect if worklets are installed or not

RNLiveMarkdown.podspec
```diff
@@ -5,8 +5,23 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
...
+end
+
+worklets_installed = is_package_installed.call('react-native-worklets')
```
tomekzaw: No need for `is_package_installed` to be a function if we only call it with `'react-native-worklets'` argument.

android/build.gradle
```diff
@@ -58,6 +58,13 @@ def resolveReactNativeDirectory() {
   throw new GradleException("[react-native-live-markdown] Unable to resolve react-native location in node_modules. Your app should define `REACT_NATIVE_NODE_MODULES_DIR` extension property in `app/build.gradle` with a path to react-native in node_modules.")
 }
 
+def isPackageInstalled = { name ->
+  def reactNativeLocation = resolveReactNativeDirectory().parent
+  if (!reactNativeLocation) return false
+    def packagePath = new File(reactNativeLocation, name)
+  return packagePath.isDirectory()
```
tomekzaw: I would prefer to use the Gradle API to check if `project(":react-native-worklets")` is found or not

src/parseExpensiMark.ts
```diff
@@ -4,10 +4,13 @@ import {Platform} from 'react-native';
 import {ExpensiMark} from 'expensify-common';
 import {unescapeText} from 'expensify-common/dist/utils';
 import {decode} from 'html-entities';
-import type {WorkletFunction} from 'react-native-reanimated/lib/typescript/commonTypes';
+import type {WorkletFunction as WorkletFunctionReanimated} from 'react-native-reanimated/lib/typescript/commonTypes';
+import type {WorkletFunction as WorkletFunctionWorklets} from 'react-native-worklets';
```
tomekzaw: Is it okay to import types from a package that might not be installed? In this case only either Reanimated or Worklets is installed

src/MarkdownTextInput.tsx
```diff
@@ -43,13 +46,45 @@ function initializeLiveMarkdownIfNeeded() {
...
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires,import/no-unresolved
+    makeShareableCloneRecursive = require('react-native-worklets').makeShareableCloneRecursive;
+  } catch {
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires,import/no-unresolved
+      makeShareableCloneRecursive = require('react-native-reanimated').makeShareableCloneRecursive;
+    } catch {
+      /* empty */
+    }
```
tomekzaw: No need for empty `catch` blocks, we actually want this to fail if the `require` fails

src/MarkdownTextInput.tsx
```diff
@@ -43,13 +46,45 @@ function initializeLiveMarkdownIfNeeded() {
...
+      createWorkletRuntime = require('react-native-reanimated').createWorkletRuntime;
+    } catch {
+      /* empty */
+    }
+  }
+
   workletRuntime = createWorkletRuntime('LiveMarkdownRuntime');
+  if (!workletRuntime) {
+    return;
+  }
```
tomekzaw: Why do we need this?

src/MarkdownTextInput.tsx
```diff
@@ -43,13 +46,45 @@ function initializeLiveMarkdownIfNeeded() {
...
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires,import/no-unresolved
+    createWorkletRuntime = require('react-native-worklets').createWorkletRuntime;
+  } catch {
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires,import/no-unresolved
+      createWorkletRuntime = require('react-native-reanimated').createWorkletRuntime;
+    } catch {
+      /* empty */
+    }
```
tomekzaw: Same as below

package.json
```diff
@@ -98,17 +98,27 @@
...
+    "react-native-reanimated": "3.19.0",
+    "react-native-worklets": "0.6.0"
   },
   "peerDependencies": {
     "expensify-common": ">=2.0.148",
     "react": "*",
     "react-native": "*",
-    "react-native-reanimated": ">=3.17.0"
+    "react-native-reanimated": ">=3.17.0",
+    "react-native-worklets": "^0.6.0"
```
tomekzaw: This is incorrect, `react-native-worklets` cannot be installed when using `react-native-reanimated@3`.

package.json
```diff
@@ -98,17 +98,27 @@
...
     "react-native": "0.81.4",
     "react-native-builder-bob": "^0.30.2",
-    "react-native-reanimated": "3.19.0",
     "react-native-web": "^0.20.0",
     "release-it": "^15.0.0",
     "turbo": "^1.10.7",
-    "typescript": "^5.8.3"
+    "typescript": "^5.8.3",
+    "react-native-reanimated": "3.19.0",
+    "react-native-worklets": "0.6.0"
```
tomekzaw: We shouldn't be using `react-native-worklets` with `react-native-reanimated@3`.

package.json
war-in: Yes, that's why I marked both as optional

https://github.com/Expensify/react-native-live-markdown/pull/731/files/2e125b8ee4268424bed20989cd73bb7f8707d6fe#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R115-R122

package.json
war-in: It's fine because those are only the dev dependencies

tomekzaw: Can we at least use Reanimated v4 instead of v3 here?

war-in:
I verified that those changes work correctly in the current E/App setup (reanimated3). Adhoc builds are here: https://github.com/Expensify/App/pull/71973

I'll continue verifying the setup with reanimated v4 and worklets

war-in:
I also confirm that E/App with `reanimatedv4` and `react-native-worklets` work fine with the code from this PR. Adhoc builds here: https://github.com/Expensify/App/pull/69469#issuecomment-3376300596

src/MarkdownTextInput.tsx
```diff
@@ -43,13 +39,33 @@ function initializeLiveMarkdownIfNeeded() {
...
 }
 
-function registerParser(parser: (input: string) => MarkdownRange[]): number {
+function registerParser(parser: (input: string) => MarkdownRange[], workletsPackageName = 'react-native-worklets'): number {
   initializeLiveMarkdownIfNeeded();
+
+  let makeShareableCloneRecursive;
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    makeShareableCloneRecursive = require(workletsPackageName).makeShareableCloneRecursive;
```
blazejkustra: I'm surprised this is allowed 😮

blazejkustra (commented):
Looks good to me (mostly reviewed the js part) 👌

src/MarkdownTextInput.tsx
war-in: yeah, I consulted it with `tjzel` (our metro expert) and it seems that this is allowed because it lives in the `try/catch` block

I tested that in the sample RN app and it worked fine (I wanted to make sure we don't have some strange metro config which allows this)

RNLiveMarkdown.podspec
```diff
@@ -5,8 +5,16 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
 react_native_minor_version = react_native_json['version'].split('.')[1].to_i
 
 pods_root = Pod::Config.instance.project_pods_root
-react_native_reanimated_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] || File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('react-native-reanimated/package.json')"`)
-react_native_reanimated_node_modules_dir_from_pods_root = Pathname.new(react_native_reanimated_node_modules_dir).relative_path_from(pods_root).to_s
+
+worklets_installed = system(%Q[
+  cd "#{Pod::Config.instance.installation_root}" &&
+  node -e "require.resolve('react-native-worklets/package.json')" > /dev/null 2>&1
+])
```
tomekzaw, line 12: This looks suspicious, in Reanimated we did it a bit differently: https://github.com/software-mansion/react-native-reanimated/blob/69422428d7eeb5414577e619296e0398048fb232/packages/react-native-reanimated/scripts/reanimated_utils.rb#L21

RNLiveMarkdown.podspec
```diff
@@ -5,8 +5,16 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
...
+package_name = worklets_installed ? 'react-native-worklets/package.json' : 'react-native-reanimated/package.json'
```
tomekzaw:
```suggestion
package_name = worklets_installed ? 'react-native-worklets' : 'react-native-reanimated'
```

RNLiveMarkdown.podspec
```diff
@@ -5,8 +5,16 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
...
+
+react_native_worklets_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] ||
+ File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('#{package_name}')"`)
```
tomekzaw:
```suggestion
 File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('#{package_name}/package.json')"`)
```

RNLiveMarkdown.podspec
```diff
@@ -5,8 +5,16 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
...
-react_native_reanimated_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] || File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('react-native-reanimated/package.json')"`)
-react_native_reanimated_node_modules_dir_from_pods_root = Pathname.new(react_native_reanimated_node_modules_dir).relative_path_from(pods_root).to_s
+
+worklets_installed = system(%Q[
+  cd "#{Pod::Config.instance.installation_root}" &&
+  node -e "require.resolve('react-native-worklets/package.json')" > /dev/null 2>&1
+])
+package_name = worklets_installed ? 'react-native-worklets/package.json' : 'react-native-reanimated/package.json'
+
+react_native_worklets_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] ||
```
tomekzaw: I'm okay with leaving only `react_native_worklets` in the variable name but we can also do it this way for now:
```suggestion
react_native_worklets_or_reanimated_node_modules_dir = ENV['REACT_NATIVE_WORKLETS_NODE_MODULES_DIR'] ||
ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] ||
```

RNLiveMarkdown.podspec
```diff
@@ -5,8 +5,16 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
...
+ File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('#{package_name}')"`)
+react_native_worklets_node_modules_dir_from_pods_root = Pathname.new(react_native_worklets_node_modules_dir).relative_path_from(pods_root).to_s
```
tomekzaw:
```suggestion
react_native_worklets_or_reanimated_node_modules_dir_from_pods_root = Pathname.new(react_native_worklets_node_modules_dir).relative_path_from(pods_root).to_s
```

android/src/main/cpp/CMakeLists.txt
```diff
@@ -4,6 +4,11 @@ cmake_minimum_required(VERSION 3.13)
 
 set(CMAKE_VERBOSE_MAKEFILE on)
 
+if(WORKLETS_INSTALLED)
```
tomekzaw, line 7: Why do we call add folly flags only if worklets are installed? Shouldn't we also do the same for Reanimated v3? Is it done this way to avoid changes in Reanimated v3 mode?

apple/MarkdownParser.mm
```diff
@@ -19,7 +19,11 @@ @implementation MarkdownParser {
     const auto &markdownRuntime = expensify::livemarkdown::getMarkdownRuntime();
     jsi::Runtime &rt = markdownRuntime->getJSIRuntime();
 
+#ifdef WORKLETS_INSTALLED
+    std::shared_ptr<SerializableWorklet> markdownWorklet;
+#else
     std::shared_ptr<ShareableWorklet> markdownWorklet;
+#endif
```
tomekzaw, line 26: How about using a typedef in a single place rather than having `#ifdef` scattered over the whole file? Something like (not the `#ifndef` instead of `#ifdef`)
```cpp
#ifndef WORKLETS_INSTALLED
using SerializableWorklet = ShareableWorklet; // ShareableWorklet was renamed to SerializableWorklet
#endif
```
This way we can use the new type name everywhere else (for Worklets compatibility) as well as the old type name (for Reanimated v3 compatibility). Except for `extractSerializableOrThrow`.

RNLiveMarkdown.podspec
```diff
@@ -24,15 +32,23 @@ Pod::Spec.new do |s|
...
-      "\"$(PODS_ROOT)/#{react_native_reanimated_node_modules_dir_from_pods_root}/apple\"",
-      "\"$(PODS_ROOT)/#{react_native_reanimated_node_modules_dir_from_pods_root}/Common/cpp\"",
+      "\"$(PODS_ROOT)/#{react_native_worklets_node_modules_dir_from_pods_root}/apple\"",
+      "\"$(PODS_ROOT)/#{react_native_worklets_node_modules_dir_from_pods_root}/Common/cpp\"",
     ].join(' '),
   }
+  if worklets_installed
+    xcconfig["OTHER_CFLAGS"] << " -DWORKLETS_INSTALLED=1"
+  end
+  s.xcconfig = xcconfig
```
tomekzaw, line 53: Does it work if we modify `s.xcconfig` directly (without `xcconfig` variable)?

RNLiveMarkdown.podspec
```diff
@@ -24,15 +32,23 @@ Pod::Spec.new do |s|
...
     "OTHER_CFLAGS" => "$(inherited) -DREACT_NATIVE_MINOR_VERSION=#{react_native_minor_version}",
     "HEADER_SEARCH_PATHS" => [
-      "\"$(PODS_ROOT)/#{react_native_reanimated_node_modules_dir_from_pods_root}/apple\"",
-      "\"$(PODS_ROOT)/#{react_native_reanimated_node_modules_dir_from_pods_root}/Common/cpp\"",
+      "\"$(PODS_ROOT)/#{react_native_worklets_node_modules_dir_from_pods_root}/apple\"",
+      "\"$(PODS_ROOT)/#{react_native_worklets_node_modules_dir_from_pods_root}/Common/cpp\"",
     ].join(' '),
   }
+  if worklets_installed
+    xcconfig["OTHER_CFLAGS"] << " -DWORKLETS_INSTALLED=1"
```
tomekzaw, line 51: On Android we don't define the value
```suggestion
    xcconfig["OTHER_CFLAGS"] << " -DWORKLETS_INSTALLED"
```

android/src/main/cpp/CMakeLists.txt
```diff
@@ -15,16 +20,32 @@ file(GLOB CPP_SRC CONFIGURE_DEPENDS "${CPP_DIR}/*.cpp")
 
 add_library(${CMAKE_PROJECT_NAME} SHARED ${ANDROID_SRC} ${CPP_SRC})
 
-target_include_directories(${CMAKE_PROJECT_NAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR} ${CPP_DIR})
+if(WORKLETS_INSTALLED)
+  target_include_directories(${CMAKE_PROJECT_NAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR} ${CPP_DIR} "${REACT_NATIVE_ROOT_DIR}/ReactCommon/jsiexecutor")
+else()
+  target_include_directories(${CMAKE_PROJECT_NAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR} ${CPP_DIR})
+endif()
```
tomekzaw, line 27: Could you please remind me why we need to include the headers from `jsiexecutor` directory?

cpp/RuntimeDecorator.cpp
```diff
@@ -23,7 +23,11 @@ void injectJSIBindings(jsi::Runtime &rt) {
       jsi::PropNameID::forAscii(rt, "jsi_registerMarkdownWorklet"),
       1,
       [](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value {
+        #ifdef WORKLETS_INSTALLED
+        auto parserId = registerMarkdownWorklet(extractSerializableOrThrow<SerializableWorklet>(rt, args[0]));
+        #else
         auto parserId = registerMarkdownWorklet(extractShareableOrThrow<ShareableWorklet>(rt, args[0]));
+        #endif
```
tomekzaw:
```suggestion
        #ifdef WORKLETS_INSTALLED
        auto worklet = extractSerializableOrThrow<SerializableWorklet>(rt, args[0]);
        #else
        auto worklet = extractShareableOrThrow<ShareableWorklet>(rt, args[0]);
        #endif
        auto parserId = registerMarkdownWorklet(worklet);
```

src/MarkdownTextInput.tsx
```diff
@@ -43,13 +39,33 @@ function initializeLiveMarkdownIfNeeded() {
...
+    createWorkletRuntime = require('react-native-reanimated').createWorkletRuntime;
+  }
   workletRuntime = createWorkletRuntime('LiveMarkdownRuntime');
+
   global.jsi_setMarkdownRuntime(workletRuntime);
   initialized = true;
 }
 
-function registerParser(parser: (input: string) => MarkdownRange[]): number {
+function registerParser(parser: (input: string) => MarkdownRange[], workletsPackageName = 'react-native-worklets'): number {
```
tomekzaw: So the trick is to pass function argument (with a default value) to `require`, is that correct?

RNLiveMarkdown.podspec
war-in: Unfortunately not, I tried that at first but there was an error about modifying s.xcconfig

src/MarkdownTextInput.tsx
war-in: Yes, that way metro can't resolve it in a build phase and only does that in runtime where error will be handled by `try/catch`

RNLiveMarkdown.podspec
tomekzaw: Okay, alternatively we can just have `other_cflags` string variable and append to it instead of having `xcconfig` variable

android/src/main/cpp/CMakeLists.txt
war-in: > Is it done this way to avoid changes in Reanimated v3 mode?

Yes, it worked fine previously so I didn't want to introduce any new changes

I took most of the code from this patch - https://github.com/Expensify/react-native-live-markdown/issues/726
Without it I got build errors

android/src/main/cpp/CMakeLists.txt
war-in: I got build errors when it was not included, took this change from this patch - https://github.com/Expensify/react-native-live-markdown/issues/726

war-in:
@tomekzaw I tested it on Expensify main and on @blazejkustra PR with reanimated bump and confirm it works as expected. 

I think we're ready for merge!

cpp/RuntimeDecorator.cpp
```diff
@@ -23,7 +23,12 @@ void injectJSIBindings(jsi::Runtime &rt) {
       jsi::PropNameID::forAscii(rt, "jsi_registerMarkdownWorklet"),
       1,
       [](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value {
-        auto parserId = registerMarkdownWorklet(extractShareableOrThrow<ShareableWorklet>(rt, args[0]));
+        #ifdef WORKLETS_INSTALLED
```
tomekzaw:
```suggestion
#ifdef WORKLETS_INSTALLED
```

cpp/RuntimeDecorator.cpp
```diff
@@ -23,7 +23,12 @@ void injectJSIBindings(jsi::Runtime &rt) {
...
+        auto worklet = extractSerializableOrThrow<SerializableWorklet>(rt, args[0]);
+        #else
```
tomekzaw:
```suggestion
#else
```

cpp/RuntimeDecorator.cpp
```diff
@@ -23,7 +23,12 @@ void injectJSIBindings(jsi::Runtime &rt) {
...
+        auto worklet = extractShareableOrThrow<SerializableWorklet>(rt, args[0]);
+        #endif
```
tomekzaw:
```suggestion
#endif
```

RNLiveMarkdown.podspec
```diff
@@ -5,8 +5,18 @@ react_native_json = JSON.parse(File.read(File.join(react_native_node_modules_dir
 react_native_minor_version = react_native_json['version'].split('.')[1].to_i
 
 pods_root = Pod::Config.instance.project_pods_root
-react_native_reanimated_node_modules_dir = ENV['REACT_NATIVE_REANIMATED_NODE_MODULES_DIR'] || File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('react-native-reanimated/package.json')"`)
-react_native_reanimated_node_modules_dir_from_pods_root = Pathname.new(react_native_reanimated_node_modules_dir).relative_path_from(pods_root).to_s
+
+worklets_installed = system(%Q[
+  cd "#{Pod::Config.instance.installation_root}" &&
+  node -e "require.resolve('react-native-worklets/package.json')" > /dev/null 2>&1
+])
```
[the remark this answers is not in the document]
war-in: that's a valid comment - I forgot to remove the previous logic of calculating `worklets_installed` value but since it has no effect now, I'll remove it in a followup PR (where we'll remove the `react-native-reanimated` entirely)

Issue #71, opened 2024-01-02
Weird things happen on Android


tomekzaw:
https://github.com/tomekzaw/react-native-markdown-text-input/assets/20516055/ce8fbd40-c6ca-43f4-b4fe-f2149235e7bc

PR #732, opened 2025-10-06
Fix applying styles for inline code syntaxes

Skalakid:
### Details

This is a follow-up PR to the previous [fix for cursor positioning in inline code markdown.](https://github.com/Expensify/react-native-live-markdown/pull/729) Here, I introduced a mechanism that marks opening and closing syntax, and based on that, it extends the range before opening syntax and the range after closing syntax. Thanks to that, styles are continuous  

### Related Issues

https://github.com/Expensify/App/issues/69493

### Manual Tests

1. Open the example app
2. Paste the following text:
```
_test🚀_
~🚀test~
~_tes🚀t_~
_`test🚀`_
~`🚀test`~
~_`tes🚀t`_~
```
3. Verify if strikethrough or italic styles aren't applied to any emoji
4. Verify I styles are properly applied on opening and closing syntaxes of inline code markdowns

### Linked PRs

https://github.com/Expensify/react-native-live-markdown/pull/729

### DIFF
--- a/src/__tests__/parseExpensiMark.test.ts
+++ b/src/__tests__/parseExpensiMark.test.ts
@@ -1,10 +1,11 @@
 import {expect} from '@jest/globals';
 import type {MarkdownRange} from '../commonTypes';
 import parseExpensiMark from '../parseExpensiMark';
+import type {ExtendedMarkdownRange} from '../rangeUtils';
 
 declare module 'expect' {
   interface Matchers<R> {
-    toBeParsedAs(expectedRanges: MarkdownRange[]): R;
+    toBeParsedAs(expectedRanges: ExtendedMarkdownRange[]): R;
   }
 }
 
@@ -222,9 +223,9 @@ test('email with multiline hyperlinks', () => {
 
 test('inline code', () => {
   expect('Hello `world`!').toBeParsedAs([
-    {type: 'syntax', start: 6, length: 1},
+    {type: 'syntax', start: 6, length: 1, syntaxType: 'opening'},
     {type: 'code', start: 7, length: 5},
-    {type: 'syntax', start: 12, length: 1},
+    {type: 'syntax', start: 12, length: 1, syntaxType: 'closing'},
   ]);
 });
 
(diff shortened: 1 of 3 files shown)


src/rangeUtils.ts
```diff
@@ -56,20 +60,22 @@ function ungroupRanges(ranges: MarkdownRange[]): MarkdownRange[] {
   });
   return ungroupedRanges;
 }
-
 /**
  * Creates a list of ranges that should not be formatted by certain markdown types (italic, strikethrough).
  * This includes emojis and syntaxes of inline code blocks.
  */
-function getRangesToExcludeFormatting(ranges: MarkdownRange[]) {
+function getRangesToExcludeFormatting(ranges: MarkdownRange[]): ExtendedMarkdownRange[] {
```
jmusial: NAB but i think this should do and you can get rid of `#71` then
```suggestion
function getRangesToExcludeFormatting(ranges: ExtendedMarkdownRange[]): ExtendedMarkdownRange[] {
```

Skalakid: If I do that, I will get eslint[no-param-reassign](https://eslint.org/docs/latest/rules/no-param-reassign). So we need to keep `#71`. Removing `as` doesn't seem to me to be worthy, because this function should transform `MarkdownRange` to `ExtendedMarkdownRange`. It's more logical to me in that way

src/__tests__/parseExpensiMark.test.ts
```diff
@@ -1,10 +1,11 @@
 import {expect} from '@jest/globals';
 import type {MarkdownRange} from '../commonTypes';
 import parseExpensiMark from '../parseExpensiMark';
+import type {ExtendedMarkdownRange} from '../rangeUtils';
```
tomekzaw: Not a huge fan of `ExtendedMarkdownRange`, why do we need this? Why can't we just extend the existing `MarkdownRange` type?

Skalakid: In fact, we can 🤔 Changed that in [0169cdc](https://github.com/Expensify/react-native-live-markdown/pull/732/commits/0169cdcbcb4c9fddea7b6254aa5b27e7308589a8)

jmusial (approved):

tomekzaw (approved):

PR #680, opened 2025-05-20
fix: update currentlyFocusedNode on focus

truph01:
### Details

- This PR implements the solution described in [this proposal](https://github.com/Expensify/App/issues/61521#issuecomment-2866029202), combined with [this suggestion](https://github.com/Expensify/App/issues/61521#issuecomment-2878112758).
- It attempts to update the `_currentlyFocusedNode` when the markdown text input receives focus.
### Related Issues

https://github.com/Expensify/App/issues/61521

### Manual Tests

1. Go to ND
2. Login account with more chats
3. Tap search
4. Scroll the page up and down
5. Verify: In search, scrolling the page must close the keypad.
### Result:

https://github.com/user-attachments/assets/4ed57083-5253-4253-80b3-943ed9737a6d

PR #736, opened 2025-10-28
Fix input loses focus when switching to another tab

bernhardoj:
### Details

This happens after this https://github.com/Expensify/react-native-live-markdown/pull/680. In that PR, we want to save the focused text input as outlined in the proposal https://github.com/Expensify/App/issues/61521#issuecomment-2866029202, so that `Keyboard.dismiss` will work on live markdown. But since we can't directly access the text input state, we do it through `RNTextInput.State.focusTextInput` on focus and `RNTextInput.State.blurTextInput` on blur.

When we switch tabs, the input `onBlur` is triggered, so the input is blurred. In normal cases, the input doesn't lose focus even though `onBlur` is called when switching tabs.

### Related Issues

Expensify: https://github.com/Expensify/App/issues/72458

### Manual Tests

1. Open Web example
2. Focus on the input
3. Switch to another tab
4. Go back to the example tab
5. Verify the input is still focused

https://github.com/user-attachments/assets/f4cf94b4-5219-419d-9f60-031deff24ad7

### DIFF
--- a/src/MarkdownTextInput.web.tsx
+++ b/src/MarkdownTextInput.web.tsx
@@ -605,7 +605,9 @@ const MarkdownTextInput = React.forwardRef<MarkdownTextInput, MarkdownTextInputP
     const handleBlur: FocusEventHandler<HTMLDivElement> = useCallback(
       (event) => {
         const e = event as unknown as NativeSyntheticEvent<TextInputFocusEventData>;
-        RNTextInput.State.blurTextInput?.(e.target);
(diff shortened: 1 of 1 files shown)


cead22:
@jmusial bump on this review please

jmusial:
@cead22  sorry, was OOO, looking now

bernhardoj:
@jmusial nope. You can merge this one.

cead22 (approved):

PR #743, opened 2025-12-11
fix: set C++ standard to C++20 in CMakeLists.txt

WoLewicki:
react-native-worklets needs c++20 from `0.7.x` cc @tjzel if I'm wrong.

### Manual Tests

Try to build live-markdown with rn-worklets `0.7.x`


tjzel:
It's true.

PR #758, opened 2026-06-25
fix(ios): persist RCTMarkdownUtils across shadow node clones to prevent AppHang

abzokhattab:
### Details

On every Yoga measure pass, `applyMarkdownFormattingToTextInputState` was allocating a **fresh** `RCTMarkdownUtils` / `MarkdownParser` instance. `MarkdownParser` has a one-entry memo cache keyed on `{text, parserId}` — because a new instance was created on each call, the cache was discarded immediately after every parse, forcing a **full JSI re-parse** of the entire input string on the main thread every time.

Yoga calls the measure callback **multiple times per layout pass**. For inputs with complex markdown, the cumulative cost of `N × full-parse` exceeded iOS's ~2 s watchdog → AppHang.

**Fix:** Add a `mutable std::shared_ptr<void> markdownUtils_` member to `MarkdownTextInputDecoratorShadowNode`:
- Lazily initialised on first use inside `applyMarkdownFormattingToTextInputState`.
- Carried through the clone constructor — when React clones a shadow node during re-render / layout, the new instance inherits the pointer, so the `RCTMarkdownUtils` (and its parser cache) is never discarded mid-layout.

Unchanged text now hits the parser's cache and skips the expensive JSI call entirely.

### Related Issues

https://github.com/Expensify/App/issues/93831

### Manual Tests

1. Paste the following comment into the composer input: https://expensify.slack.com/archives/C05LX9D6E07/p1782316444025709?thread_ts=1782312343.503839&cid=C05LX9D6E07
2. Open and close the keyboard.
3. Verify that the app doesn't crash.

https://github.com/user-attachments/assets/5f1623b4-5e0b-4472-b984-1957f83d4258

https://github.com/user-attachments/assets/86c13fef-f6cc-4aaf-9c9a-3bd595ee238b

### DIFF
--- a/apple/MarkdownTextInputDecoratorShadowNode.mm
+++ b/apple/MarkdownTextInputDecoratorShadowNode.mm
@@ -32,6 +32,13 @@ MarkdownTextInputDecoratorShadowNode::MarkdownTextInputDecoratorShadowNode(
     ShadowNode const &sourceShadowNode,
     ShadowNodeFragment const &fragment)
     : ConcreteViewShadowNode(sourceShadowNode, fragment) {
+  // Carry the persisted RCTMarkdownUtils over from the source node so the
+  // MarkdownParser memo cache survives the frequent cloning that happens
+  // during layout and re-render cycles.
+  const auto &source =
+      static_cast<const MarkdownTextInputDecoratorShadowNode &>(sourceShadowNode);
+  markdownUtils_ = source.markdownUtils_;
+
   initialize();
   makeChildNodeMutable();
   
@@ -182,10 +189,19 @@ void MarkdownTextInputDecoratorShadowNode::applyMarkdownFormattingToTextInputSta
   const auto defaultNSTextAttributes =
       RCTNSTextAttributesFromTextAttributes(defaultTextAttributes);
 
-  // this can possibly be optimized
+  // Lazily create and persist the RCTMarkdownUtils instance so the MarkdownParser
+  // one-entry memo cache (keyed on text + parserId) survives repeated Yoga measure
+  // callbacks. Previously a fresh utils/parser was allocated on every call,
+  // discarding the cache and forcing a full JSI re-parse each time.
+  if (!markdownUtils_) {
+    RCTMarkdownUtils *freshUtils = [[RCTMarkdownUtils alloc] init];
+    markdownUtils_ = std::shared_ptr<void>(
+        (__bridge_retained void *)freshUtils, [](void *p) { CFRelease(p); });
+  }
+  RCTMarkdownUtils *utils = (__bridge RCTMarkdownUtils *)markdownUtils_.get();
+
   RCTMarkdownStyle *markdownStyle =
       [[RCTMarkdownStyle alloc] initWithStruct:decoratorProps.markdownStyle];
-  RCTMarkdownUtils *utils = [[RCTMarkdownUtils alloc] init];
   [utils setMarkdownStyle:markdownStyle];
   [utils setParserId:[NSNumber numberWithInt:decoratorProps.parserId]];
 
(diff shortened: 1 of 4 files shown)


abzokhattab:

apple/MarkdownTextInputDecoratorShadowNode.mm
war-in, line 1: I only worry that having one `markdownUtils` instance can introduce unexpected errors. I wonder if we create it once it won't cache too much and we end up having stale data in some input (e.g. if using multiple)

@abzokhattab could you verify if this solution works well with two inputs defined on the same page? 🙏  We have an example app here in live-markdown repo so it should be fairly easy to setup

abzokhattab: Added two inputs in the example app, and it's working as expected

https://github.com/user-attachments/assets/86c13fef-f6cc-4aaf-9c9a-3bd595ee238b

situchan:
Works for me

https://github.com/user-attachments/assets/fe45f8d3-207f-4d07-88d0-beb3d95d3e0f

apple/MarkdownTextInputDecoratorShadowNode.mm
```diff
@@ -182,10 +189,19 @@
...
+    markdownUtils_ = std::shared_ptr<void>(
+        (__bridge_retained void *)freshUtils, [](void *p) { CFRelease(p); });
+  }
+  RCTMarkdownUtils *utils = (__bridge RCTMarkdownUtils *)markdownUtils_.get();
+
   RCTMarkdownStyle *markdownStyle =
       [[RCTMarkdownStyle alloc] initWithStruct:decoratorProps.markdownStyle];
-  RCTMarkdownUtils *utils = [[RCTMarkdownUtils alloc] init];
   [utils setMarkdownStyle:markdownStyle];
   [utils setParserId:[NSNumber numberWithInt:decoratorProps.parserId]];
```
situchan: lazy-init block + the `setMarkdownStyle:`/`setParserId:` calls right after) — sharing one mutable `RCTMarkdownUtils` across all clones in a family re-introduces cross-clone state coupling that the previous fresh-per-call instance was immune to.**

`setMarkdownStyle:` and `setParserId:` mutate the shared `RCTMarkdownUtils`/`MarkdownParser` *outside* the parser's `@synchronized(self)` block, which only covers `parse:`.
If two clones of the same family are measured concurrently on different threads (e.g. overlapping background layout vs. a synchronous main-thread text measurement during mount), clone A's `setParserId:`/`setMarkdownStyle:` can interleave with clone B's, so A's `parse:`/`formatAttributedString:` runs against B's `parserId`/style — producing wrong markdown ranges or wrong styling for that frame.
With the old fresh-instance-per-call code this interleaving was impossible.
The trigger (concurrent measure of the same family) is uncommon and the result is transient, not a crash — hence PLAUSIBLE rather than CONFIRMED.

So I'd like to confirm that decorator measurement is single-threaded per family; if so, this is a non-issue and the fix is clean as-is.

situchan (approved):

apple/MarkdownTextInputDecoratorShadowNode.mm
abzokhattab: Good catch .. confirmed it's reachable. In RN 0.85.3, `ShadowTree::tryCommit` runs `layoutIfNeeded` (→ our measure callback) **lock-free** between a shared-lock read and a unique-lock CAS (with retry), and Yoga layout uses a `threadLocalLayoutContext` — so commits/layout run concurrently across threads. Clones share the same `markdownUtils_`, and `setMarkdownStyle:`/`setParserId:` were outside `parse:`'s lock, so they could interleave and apply the wrong style/parserId for a frame (transient, non-crashing).

Fixed in the last commit: added an `applyMarkdownFormatting:…markdownStyle:parserId:` overload that sets style/parserId + parse + format under one `@synchronized(self)` block (recursive, nests with `parse:`'s lock); the measure path now uses it. The view-layer instance is separate/main-thread, so unaffected.

 

https://github.com/user-attachments/assets/a90fe972-5123-4486-80ca-1c956c3a673e

abzokhattab:
cc @mountiny

mountiny:
Thanks, looking for someone from SWM to do a final review today as well

mountiny (approved):

jmusial:
hey @war-in is OOO, I'm gonna check it soon

jmusial:
LGTM!

PR #763, opened 2026-07-03
Bump react-native-worklets to 0.10.x

tomekzaw:
### Details
Bumps `react-native-worklets` to `0.10.1` and adds compatibility with the 0.10.x line.

Worklets 0.10 no longer silently substitutes an inaccessible placeholder for non-plain objects it can't serialize — it throws `` Cannot copy value of type `Logger` `` instead. The built-in `parseExpensiMark` parser's worklet closure transitively captures the `Log` singleton from `expensify-common` (a `Logger` class instance), so registering the parser crashes the moment a `<MarkdownTextInput>` mounts.

The logger is never invoked on the worklet runtime, so we register a custom serializer (`registerCustomSerializable`) that replaces any `Logger` with a no-op stand-in on the worklet side.

### Related Issues
N/A

### Manual Tests
- Ran the example app on the iOS simulator: without the fix it renders the `` Cannot copy value of type `Logger` `` red screen on mount; with the fix the editor renders and live Markdown formatting works (typed `*live*` → bold).

### Linked PRs
N/A

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@expensify/react-native-live-markdown",
-  "version": "0.1.331",
+  "version": "0.1.332",
   "description": "Drop-in replacement for React Native's TextInput component with Markdown formatting.",
   "main": "lib/commonjs/index",
   "module": "lib/module/index",


tomekzaw:
Here's the error that shows up if we only bump `react-native-worklets` to 0.10.x:

<table>
<thead>
<tr>
<th>Android</th>
<th>iOS</th>
</tr>
</thead>
<tbody>
<tr>
<td>

</td>
<td>

</td>
</tr>
</tbody>
</table>

j-piasecki (approved):

PR #769, opened 2026-07-21
feat: live-markdown working in both Bundle Mode and Legacy Eval Mode

tjzel:
### Details

I made changes necessary for the library to be working in both modes.

### Related Issues

n/a

### Manual Tests

I ran the app in both modes and it works - there's also a mode display text in the example.

### Linked PRs

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,23 @@ cd ios && bundler install && bundler exec pod install
 
 The library includes native code so you will need to re-build the native app.
 
+### Worklets Bundle Mode
+
+If you're using [Worklets Bundle Mode](https://docs.swmansion.com/react-native-worklets/docs/bundleMode) you will have to
+configure your [import forwarding](https://docs.swmansion.com/react-native-worklets/docs/worklets-babel-plugin/plugin-options#importforwarding) as follows in your `babel.config.js` file:
+
+```js
+importForwarding: {
+  moduleNames: [
+    // your existing module names here...
+    // the order of the modules does not matter
+    'expensify-common',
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -33,6 +33,23 @@ cd ios && bundler install && bundler exec pod install
 
 The library includes native code so you will need to re-build the native app.
 
+### Worklets Bundle Mode
+
+If you're using [Worklets Bundle Mode](https://docs.swmansion.com/react-native-worklets/docs/bundleMode) you will have to
+configure your [import forwarding](https://docs.swmansion.com/react-native-worklets/docs/worklets-babel-plugin/plugin-options#importforwarding) as follows in your `babel.config.js` file:
```
war-in, line 39: Let's also redirect users to the correct section of the Worklets docs about their metro setup (metro.config.js, and the patches). The current note suggests that they only have to configure import forwarding

example/package.json
```diff
@@ -6,7 +6,8 @@
     "android": "react-native run-android",
     "ios": "react-native run-ios",
     "lint": "eslint .",
-    "start": "react-native start --client-logs"
+    "start": "react-native start --client-logs",
+    "bundle-mode-start": "BUNDLE_MODE=1 react-native start --reset-cache --client-logs"
```
war-in, line 10: I wonder if we shouldn't make it the default 🤔 Are there any drawbacks to always enabling bundle mode?

war-in (commented):
Thank you for the contribution! I left two small comments and once addressed we can merge it

PR #772, opened 2026-08-10
Fix main thread hang on iOS

staszekscp:
### Details

On iOS, `RCTMarkdownUtils` and `MarkdownParser` nest an Objective-C lock *outside* the markdown worklet runtime's `recursive_mutex`. That ordering lets a background thread hold the ObjC lock while it waits on the runtime mutex, which blocks the main thread in `objc_sync_enter` during Yoga measure until the iOS watchdog kills the app.

Today's code:

```objc
// RCTMarkdownUtils.mm
@synchronized (self) {
  _markdownStyle = markdownStyle;
  _parserId = parserId;
  [self applyMarkdownFormatting:...];   // -> MarkdownParser parse: -> runSync into the runtime
}

// MarkdownParser.mm
@synchronized (self) {
  ...
  output = markdownRuntime->runGuarded(markdownWorklet, input);   // blocks on recursive_mutex
  ...
}
```

Both locks are taken before entering the runtime, and the parse is synchronous, so the lock is held for the entire duration of the worklet call.

**The two threads in the captured hang:**

`com.facebook.react.runtime.JavaScript` — holds both `@synchronized` locks, blocked on the runtime mutex:

```
facebook::yoga::Node::measure
MarkdownTextInputDecoratorShadowNode::yogaNodeMeasureCallbackConnector
MarkdownTextInputDecoratorShadowNode::measureContent
MarkdownTextInputDecoratorShadowNode::applyMarkdownFormattingToTextInputState
-[RCTMarkdownUtils applyMarkdownFormatting:withDefaultTextAttributes:markdownStyle:parserId:]
-[RCTMarkdownUtils applyMarkdownFormatting:withDefaultTextAttributes:]
-[MarkdownParser parse:withParserId:]
worklets::WorkletRuntime::runSync<T>
...
worklets::WorkletRuntime::runSyncSerialized<T>
std::__1::recursive_mutex::lock
_pthread_mutex_firstfit_lock_slow
__psynch_mutexwait            <-- blocked here
```

Main thread — same measure path, blocked acquiring the lock the thread above is holding:

```
facebook::react::YogaLayoutableShadowNode::layoutTree
...
facebook::yoga::Node::measure
MarkdownTextInputDecoratorShadowNode::yogaNodeMeasureCallbackConnector
MarkdownTextInputDecoratorShadowNode::measureContent
MarkdownTextInputDecoratorShadowNode::applyMarkdownFormattingToTextInputState
-[RCTMarkdownUtils applyMarkdownFormatting:withDefaultTextAttributes:markdownStyle:parserId:]
objc_sync_enter
_os_unfair_lock_lock_slow
__ulock_wait2                 <-- blocked here
```

`Fatal App Hang Fully Blocked` — main thread blocked ≥ 2000 ms, culprit `-[RCTMarkdownUtils applyMarkdownFormatting:withDefaultTextAttributes:markdownStyle:parserId:]`. 

### The fix

Two changes, both aimed at removing "main thread waits on a runtime-bound lock":

**1. `MarkdownParser` — never hold an ObjC lock across the runtime call.**

`parse:` is split into a cache lookup, an unlocked `parseUncached:` that enters the runtime, and a locked memo write. The lock now only guards the `_prev*` ivars, so it is never held while waiting on `recursive_mutex` and the inversion is gone. Two concurrent misses may parse the same text twice; the runtime serializes them, results are identical, and last-writer-wins on a single-entry memo is safe.

Two new entry points:

- `cachedRangesForText:withParserId:` — memo lookup only, never touches the runtime, safe from the measure path on the main thread.
- `warmCacheAsyncForText:withParserId:` — schedules a parse on a serial `QOS_CLASS_USER_INITIATED` queue, guarded by an `_asyncParseInFlight` flag so warm-ups don't pile up.

**2. `RCTMarkdownUtils` — main thread formats from cache only.**

The lock narrows to the `_markdownStyle`/`_parserId` writes; formatting then uses the local parameters rather than re-reading the ivars, which preserves the per-call consistency the original single lock was there to provide. On a cache miss:

- **main thread**: schedule an async warm-up and return without formatting, so the runtime is never entered from the watchdog-monitored thread.
- **background layout threads**: parse synchronously as before. This is now safe because `parse:` no longer holds any lock the main-thread measure path can block on, and the watchdog only monitors the main thread.

### Tradeoffs / things I'd like reviewer input on

Flagging these because they're real and I'd rather they get caught here:

1. **A main-thread cache miss measures unformatted text.** Styles that change metrics (`h1` font size, `pre`/`code` font, blockquote indent, emoji font size) would measure wrong for that pass. In practice the main thread should almost always hit the cache, because text changes are committed and parsed on background threads first — but "almost always" is doing some work in that sentence.

2. **Nothing explicitly re-measures after the async warm-up lands.** The comment says a subsequent measure/commit picks up the cached ranges, and that's true whenever something else dirties layout, but there's no invalidation triggered by the warm-up itself. If Fabric has cached the measurement for that shadow node, a wrong height could persist until the text changes again. Should `warmCacheAsyncForText:` complete by dirtying the shadow node / requesting a new commit? That would make convergence guaranteed rather than incidental.

3. **The memo is single-entry, and now shared with a background writer.** Two alternating text values thrash it, and in that state the main thread would rarely hit the cache — falling into case 1 repeatedly. A small LRU (even 2–4 entries) keyed on `(text, parserId)` would make the main-thread fast path much more reliable. Happy to add it here if you'd prefer.

4. **`_asyncParseInFlight` can skip the newest text.** If a warm-up is already running for stale text, the newer text's warm-up is dropped and relies on a later measure pass observing the miss. Combined with (2), there's a path where it doesn't converge promptly. Coalescing to "latest requested text" instead of a boolean would be tighter.

An alternative worth considering: keep formatting synchronous everywhere and instead make the runtime entry non-blocking, or ensure the parse always happens on commit (before measure) so the measure path is cache-only by construction. That's a larger change; this PR is scoped to stopping the watchdog kills.

### Related Issues

https://github.com/Expensify/App/issues/93831

Internal: Sentry `APP-EF1` (https://expensify.sentry.io/issues/APP-EF1)

### Manual Tests

There is no ObjC test harness in this repo, so this was verified manually. Verified in the Expensify app (HybridApp, iOS) carrying this as a `patch-package` patch over `@expensify/react-native-live-markdown@0.1.334`:

- Markdown formatting in the composer still applies correctly — bold, italic, strikethrough, `code`, `pre`, blockquote, headings, mentions, links, emoji.
- Typing, fast typing, and paste of a large markdown block: formatting applies without a visible unformatted frame.
- Editing an existing message with markdown: correct formatting on open.
- Switching between reports with different composer drafts: no stale or cross-contaminated formatting (i.e. no wrong `parserId` applied).
- No main-thread hangs on the previously reproducing path.

Reviewers: please pay particular attention to composer height on first render of a heavily formatted draft — that's where tradeoff (1)/(2) above would show up as a layout jump or wrong height.

### Linked PRs

- Expensify/App: https://github.com/Expensify/App/pull/97871

### DIFF
--- a/apple/MarkdownParser.h
+++ b/apple/MarkdownParser.h
@@ -8,26 +8,22 @@ NS_ASSUME_NONNULL_BEGIN
 - (NSArray<MarkdownRange *> *)parse:(nonnull NSString *)text
                        withParserId:(nonnull NSNumber *)parserId;
 
-// Returns the memoized ranges for (text, parserId) if they are still in the
-// cache, otherwise nil. Never enters the worklet runtime and only takes a brief
-// internal lock, so it is safe to call from the Yoga measure path on the main
-// thread.
+// Returns the ranges for (text, parserId) if they are already in the cache, or
+// nil if they aren't. This never runs the parser, so it is safe to call from
+// the main thread while Yoga is measuring.
 - (nullable NSArray<MarkdownRange *> *)cachedRangesForText:(nonnull NSString *)text
                                               withParserId:(nonnull NSNumber *)parserId;
 
-// Requests a parse on a background queue to populate the cache without blocking
-// the calling thread. Used when the main thread needs ranges during layout but
-// must not wait on the worklet runtime (see Sentry APP-EF1).
+// Parses in the background and puts the result in the cache, so the caller
+// doesn't have to wait for the parser. Used by the main thread, which must
+// never wait for it (see Sentry APP-EF1).
 //
-// Requests are coalesced latest-wins: the most recent (text, parserId) always
-// wins, and any earlier request that has not started executing yet is dropped.
-// A parse that is already inside the worklet runtime cannot be cancelled, but
-// the newest request is picked up as soon as it returns, so the newest text is
-// always parsed.
+// Only the newest request matters: a new call replaces one that is still
+// waiting. A parse that already started can't be stopped, but the newest text
+// is parsed as soon as it finishes.
 //
-// `completion` runs on the warm-up queue once `text` has been parsed and
-// cached. It is skipped when the request was superseded by a newer one, since
-// the newer request invokes its own completion.
+// `completion` runs on the background queue once the text is cached. It is
+// skipped if a newer call replaced this one, since that call reports instead.
 - (void)warmCacheAsyncForText:(nonnull NSString *)text
                  withParserId:(nonnull NSNumber *)parserId
                    completion:(nullable void (^)(void))completion;
--- a/apple/MarkdownParser.h
+++ b/apple/MarkdownParser.h
@@ -8,6 +8,19 @@ NS_ASSUME_NONNULL_BEGIN
 - (NSArray<MarkdownRange *> *)parse:(nonnull NSString *)text
                        withParserId:(nonnull NSNumber *)parserId;
 
+// Returns the memoized ranges for (text, parserId) if they match the last
+// completed parse, otherwise nil. Never enters the worklet runtime and only
+// takes a brief internal lock, so it is safe to call from the Yoga measure
+// path on the main thread.
+- (nullable NSArray<MarkdownRange *> *)cachedRangesForText:(nonnull NSString *)text
+                                              withParserId:(nonnull NSNumber *)parserId;
+
+// Schedules a parse on a background queue to populate the memo cache without
+// blocking the calling thread. Used when the main thread needs ranges during
+// layout but must not wait on the worklet runtime (see Sentry APP-EF1).
+- (void)warmCacheAsyncForText:(nonnull NSString *)text
+                 withParserId:(nonnull NSNumber *)parserId;
+
 NS_ASSUME_NONNULL_END
 
 @end
--- a/apple/MarkdownParser.h
+++ b/apple/MarkdownParser.h
@@ -8,19 +8,30 @@ NS_ASSUME_NONNULL_BEGIN
 - (NSArray<MarkdownRange *> *)parse:(nonnull NSString *)text
                        withParserId:(nonnull NSNumber *)parserId;
 
-// Returns the memoized ranges for (text, parserId) if they match the last
-// completed parse, otherwise nil. Never enters the worklet runtime and only
-// takes a brief internal lock, so it is safe to call from the Yoga measure
-// path on the main thread.
+// Returns the memoized ranges for (text, parserId) if they are still in the
+// cache, otherwise nil. Never enters the worklet runtime and only takes a brief
+// internal lock, so it is safe to call from the Yoga measure path on the main
+// thread.
 - (nullable NSArray<MarkdownRange *> *)cachedRangesForText:(nonnull NSString *)text
                                               withParserId:(nonnull NSNumber *)parserId;
 
-// Schedules a parse on a background queue to populate the memo cache without
-// blocking the calling thread. Used when the main thread needs ranges during
-// layout but must not wait on the worklet runtime (see Sentry APP-EF1).
+// Requests a parse on a background queue to populate the cache without blocking
+// the calling thread. Used when the main thread needs ranges during layout but
+// must not wait on the worklet runtime (see Sentry APP-EF1).
+//
+// Requests are coalesced latest-wins: the most recent (text, parserId) always
+// wins, and any earlier request that has not started executing yet is dropped.
+// A parse that is already inside the worklet runtime cannot be cancelled, but
+// the newest request is picked up as soon as it returns, so the newest text is
+// always parsed.
+//
+// `completion` runs on the warm-up queue once `text` has been parsed and
+// cached. It is skipped when the request was superseded by a newer one, since
+// the newer request invokes its own completion.
 - (void)warmCacheAsyncForText:(nonnull NSString *)text
-                 withParserId:(nonnull NSNumber *)parserId;
-
-NS_ASSUME_NONNULL_END
+                 withParserId:(nonnull NSNumber *)parserId
+                   completion:(nullable void (^)(void))completion;
 
 @end
+
+NS_ASSUME_NONNULL_END
--- a/apple/MarkdownTextInputDecoratorShadowNode.mm
+++ b/apple/MarkdownTextInputDecoratorShadowNode.mm
@@ -3,8 +3,12 @@
 #include <React/RCTUtils.h>
 #include <react/renderer/components/view/conversions.h>
 #include <react/renderer/core/ComponentDescriptor.h>
+#include <react/renderer/core/ConcreteState.h>
 #include <react/renderer/textlayoutmanager/RCTAttributedTextUtils.h>
 #include <yoga/Yoga.h>
+
+#include <atomic>
+#include <memory>
 
 #include "RCTMarkdownStyle.h"
 #include "RCTMarkdownUtils.h"
@@ -33,11 +37,14 @@ MarkdownTextInputDecoratorShadowNode::MarkdownTextInputDecoratorShadowNode(
     ShadowNodeFragment const &fragment)
     : ConcreteViewShadowNode(sourceShadowNode, fragment) {
   // Carry the persisted RCTMarkdownUtils over from the source node so the
-  // MarkdownParser memo cache survives the frequent cloning that happens
-  // during layout and re-render cycles.
+  // MarkdownParser cache survives the frequent cloning that happens during
+  // layout and re-render cycles. The re-measure flag has to travel with it,
+  // and both must be in place before makeChildNodeMutable() below, which is
+  // what consumes the flag.
   const auto &source =
       static_cast<const MarkdownTextInputDecoratorShadowNode &>(sourceShadowNode);
   markdownUtils_ = source.markdownUtils_;
+  needsRemeasure_ = source.needsRemeasure_;
 
   initialize();
   makeChildNodeMutable();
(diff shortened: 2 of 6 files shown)


war-in:
Summary of the PR. Please correct me if I misunderstood something

> The main thread will always trigger the markdown parser that parses a text on a separate worklet thread, so the main thread is not blocked. This main thread won't synchronise on parsing, so in some situations the text input height can be outdated and will jump in the next commit (after parsing is finished).

Regarding the questions
1. That's a real concern. We should make sure to test that thoroughly, so we dont introduce regressions if possible
2. I'd definitely trigger a rerender if possible. We could use dirty state if necessary
3. Yes, let's save more than just one cache entry 👍 
4. I think we could start a new parse immediately and remove the stale, still ongoing one, so we're sure the newest text will always appear

staszekscp:
Hey! I think the questions from the PR description were covered according to your answers @war-in!

war-in:
That's great! Could we link this PR as a git dependency in your E/App PR, so C+ could jump in and test that solution in the Expensify app? We should make sure we won't introduce any regressions 🙏 
cc @staszekscp

staszekscp:
The changes can be tested via [this PR](https://github.com/Expensify/App/pull/97871)!

apple/RCTMarkdownUtils.mm
```diff
@@ -39,17 +39,59 @@ - (void)applyMarkdownFormatting:(nonnull NSMutableAttributedString *)attributedS
...
+        // Runs on the parser's warm-up queue, only for the newest requested
+        // text. Cannot spin: the re-measure it asks for hits the cache and stops
+        // scheduling warm-ups (or the text changed again, in which case the new
+        // text needs a re-measure anyway).
+        void (^handler)(void) = weakSelf.onAsyncFormattingReady;
+        if (handler != nil) {
+          handler();
+        }
+      }];
+      return;
```
situchan, line 78: This early returns before `MarkdownFormatter` runs whenever the measure pass is on the main thread and the parser cache misses.
Before, `parse:` was synchronous, so the string handed to `setStateData` (`MarkdownTextInputDecoratorShadowNode.mm:308, :323`) was always formatted.
`formatAttributedString` is what applies h1 font size, code/pre font, blockquote indent and emoji size — so skipping it changes the measured height.
The first measure after an input mounts is a guaranteed miss: `markdownUtils_` (and its empty `MarkdownParser` cache) is created one line earlier at `:224`.


## Repro steps

**Example app:**
1. The input is pre-filled with
```
# header1`, `> blockquote`, ```codeblock```, 😀🍕🍔
```
Note how tall the input box is and that all of it is visible inside the border.
2. Rotate the device to landscape.
3. Rotate back to portrait.
Expected: the box stays sized to the formatted text through both rotations.
Actual: on the rotation the box is briefly sized to plain-text metrics — the header line and code block lose their extra height, so the bottom lines sit outside or flush against the grey border before it snaps back.

**E/App:**
1. On iPad, open a chat and paste into the composer:
```
# heading
```code```
😀😀😀

```
2. Put the app in Split View and drag the divider, or rotate the iPad.
3. The composer collapses toward single-line height for a frame before correcting.

apple/MarkdownTextInputDecoratorShadowNode.mm
```diff
@@ -97,6 +104,26 @@
...
+  // Dirty both nodes explicitly instead of relying on
+  // YGNodeMarkDirty()'s upward propagation: the child is usually already dirty
+  // from completeClone(), in which case propagation short-circuits and the
+  // decorator would stay clean. Ancestors then pick the dirty flag up on their
+  // own, because they are cloned with new children and updateYogaChildren()
+  // propagates child dirtiness upwards.
+  if (needsRemeasure_ != nullptr && needsRemeasure_->exchange(false)) {
+    yogaNode->setDirty(true);
+    yogaNode_.setDirty(true);
+  }
```
situchan: The wrong height stays wrong instead of snapping back

**Repro step:** (example app)

1. Tap the input and start typing continuously into the middle of the text (each keystroke is a commit, which is what makes the recovery commit lose the revision race).
2.. **While still typing**, rotate the device to landscape.
3. Stop typing and look at the box.
4. If it recovered, tap **Blur**, rotate back, and repeat from step 1. It typically takes a handful of tries.

Expected: box returns to formatted height once rotation settles.
Actual: the box stays at the plain-text height with the last lines clipped, and stays there indefinitely. It only corrects on the next keystroke (which changes the text and forces a fresh measure). Blur/focus, scrolling, and further rotation do not fix it.

situchan (commented):
From AI analysis:

apple/RCTMarkdownUtils.mm
staszekscp: `main`

https://github.com/user-attachments/assets/7cfe1001-8748-432b-ba72-734fb218a1fb

staszekscp: This branch 

https://github.com/user-attachments/assets/ca09d04e-7146-4b71-bf19-1520c49d7858

staszekscp: Apparently the behaviour is exactly the same and there is no glitch

apple/MarkdownTextInputDecoratorShadowNode.mm
staszekscp: Can't see any regression:

https://github.com/user-attachments/assets/b217ee85-ce0b-4cf5-91c1-d9c525e27fb7

staszekscp:
I've verified the AI comments and couldn't see any regression that was pointed out as potential

situchan:
## Reviewer Checklist

- [x] I have verified the author checklist is complete (all boxes are checked off).
- [x] I verified the correct issue is linked in the `### Fixed Issues` section above
- [x] I verified testing steps are clear and they cover the changes made in this PR
    - [x] I verified the steps for local testing are in the `Tests` section
    - [x] I verified the steps for Staging and/or Production testing are in the `QA steps` section
    - [x] I verified the steps cover any possible failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
- [x] I checked that screenshots or videos are included for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I verified that the composer does not automatically focus or open the keyboard on mobile unless explicitly intended. This includes checking that returning the app from the background does not unexpectedly open the keyboard.
- [x] I verified tests pass on **all platforms** & I tested again on:
    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
- [x] If there are any errors in the console that are unrelated to this PR, I either fixed them (preferred) or linked to where I reported them in Slack
- [x] I verified proper code patterns were followed (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
    - [x] I verified that any callback methods that were added or modified are named for what the method does and never what callback they handle (i.e. `toggleReport` and not `onIconClick`).
    - [x] I verified that comments were added to code that is not self explanatory
    - [x] I verified that any new or modified comments were clear, correct English, and explained "why" the code was doing something instead of only explaining "what" the code was doing.
    - [x] I verified any copy / text that was added to the app is grammatically correct in English. It adheres to proper capitalization guidelines (note: only the first word of header/labels should be capitalized), and is either coming verbatim from figma or has been approved by marketing (in order to get marketing approval, ask the Bug Zero team member to add the Waiting for copy label to the issue)
- [x] If a new code pattern is added I verified it was agreed to be used by multiple Expensify engineers
- [x] I verified that this PR follows the guidelines as stated in the [Review Guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md)
- [x] I verified other components that can be impacted by these changes have been tested, and I retested again (i.e. if the PR modifies a shared library or component like `Avatar`, I verified the components using `Avatar` have been tested & I retested again)
- [x] If a new component is created I verified that:
    - [x] A similar component doesn't exist in the codebase
    - [x] All props are defined accurately and each prop has a `/** comment above it */`
    - [x] The file is named correctly
    - [x] The component has a clear name that is non-ambiguous and the purpose of the component can be inferred from the name alone
    - [x] The only data being stored in the state is data necessary for rendering and nothing else
    - [x] For Class Components, any internal methods passed to components event handlers are bound to `this` properly so there are no scoping issues (i.e. for `onClick={this.submit}` the method `this.submit` should be bound to `this` in the constructor)
    - [x] Any internal methods bound to `this` are necessary to be bound (i.e. avoid `this.submit = this.submit.bind(this);` if `this.submit` is never passed to a component event handler like `onClick`)
    - [x] All JSX used for rendering exists in the render method
    - [x] The component has the minimum amount of code necessary for its purpose, and it is broken down into smaller components in order to separate concerns and functions
- [x] If any new file was added I verified that:
    - [x] The file has a description of what it does and/or why is needed at the top of the file if the code is not self explanatory
- [x] If a new CSS style is added I verified that:
    - [x] A similar style doesn't already exist
    - [x] The style can't be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/utils/index.ts) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(theme.componentBG`)
- [x] If the PR modifies code that runs when editing or sending messages, I tested and verified there is no unexpected behavior for all supported markdown - URLs, single line code, code blocks, quotes, headings, bold, strikethrough, and italic.
- [x] If the PR modifies a generic component, I tested and verified that those changes do not break usages of that component in the rest of the App (i.e. if a shared library or component like `Avatar` is modified, I verified that `Avatar` is working as expected in all cases)
- [x] If the PR modifies a component related to any of the existing Storybook stories, I tested and verified all stories for that component are still working as expected.
- [x] If the PR modifies a component or page that can be accessed by a direct deeplink, I verified that the code functions as expected when the deeplink is used - from a logged in and logged out account.
- [x] If the PR modifies the UI (e.g. new buttons, new UI components, changing the padding/spacing/sizing, moving components, etc) or modifies the form input styles:
    - [x] I verified that all the inputs inside a form are aligned with each other.
    - [x] I added `Design` label and/or tagged `@Expensify/design` so the design team can review the changes.
- [x] For any bug fix or new feature in this PR, I verified that sufficient [unit tests](https://github.com/Expensify/App/blob/main/tests/README.md) are included to prevent regressions in this flow.
- [x] If the `main` branch was merged into this PR after a review, I tested again and verified the outcome was still expected according to the `Test` steps.
- [x] I have checked off every checkbox in the PR reviewer checklist, including those that don't apply to this PR.

### Screenshots/Videos
<details>
<summary>Android: HybridApp</summary>



</details>

<details>
<summary>Android: mWeb Chrome</summary>



</details>

<details>
<summary>iOS: HybridApp</summary>



https://github.com/user-attachments/assets/325ca98b-dfc7-4589-bd2a-584a13e6fdfc


https://github.com/user-attachments/assets/3a8d2ec9-bbb0-4fc3-aa71-887fbd50fcf5


</details>

<details>
<summary>iOS: mWeb Safari</summary>



</details>

<details>
<summary>MacOS: Chrome / Safari</summary>



</details>

apple/MarkdownParser.h
```diff
@@ -8,6 +8,30 @@ NS_ASSUME_NONNULL_BEGIN
...
+// internal lock, so it is safe to call from the Yoga measure path on the main
+// thread.
+- (nullable NSArray<MarkdownRange *> *)cachedRangesForText:(nonnull NSString *)text
+                                              withParserId:(nonnull NSNumber *)parserId;
+
+// Requests a parse on a background queue to populate the cache without blocking
+// the calling thread. Used when the main thread needs ranges during layout but
+// must not wait on the worklet runtime (see Sentry APP-EF1).
+//
+// Requests are coalesced latest-wins: the most recent (text, parserId) always
```
situchan: Let's update and simplify all the comments to be more human readable.
i.e. coalesced

situchan (approved):

mountiny (approved):
SWM and C+ approved

PR #775, opened 2026-09-03
refactor: add optional max length to parse expensimark

Uzaifm127:
### Details

`parseExpensiMark()` uses `ExpensiMark` from `expensify-common` to convert Markdown into formatted ranges.

This PR makes the parser length configurable while keeping the existing 4,000-character limit as the default by adding an optional argument to `parseExpensiMark` function. It also updates `expensify-common` to `2.0.201`, which contains the optimized `ExpensiMark` implementation required by this parser.

### Related Issues

https://github.com/Expensify/App/issues/95210

### Manual Tests

N/A

### DIFF
--- a/src/parseExpensiMark.ts
+++ b/src/parseExpensiMark.ts
@@ -238,6 +238,13 @@ function parseTreeToTextAndRanges(tree: StackItem): [string, MarkdownRange[]] {
   return [text, ranges];
 }
 
+/**
+ * Parses Markdown text into formatting ranges using ExpensiMark.
+ *
+ * @param markdown - The raw Markdown string to parse.
+ * @param maxLength - Maximum input length to parse; longer input returns `[]` without parsing. Defaults to 4000.
+ * @returns The list of formatting ranges, or `[]` if input exceeds `maxLength` or fails to parse.
+ */
 function parseExpensiMark(markdown: string, maxLength = MAX_PARSABLE_LENGTH): MarkdownRange[] {
   if (markdown.length > maxLength) {
     return [];
--- a/src/parseExpensiMark.ts
+++ b/src/parseExpensiMark.ts
@@ -238,8 +238,8 @@ function parseTreeToTextAndRanges(tree: StackItem): [string, MarkdownRange[]] {
   return [text, ranges];
 }
 
-function parseExpensiMark(markdown: string): MarkdownRange[] {
-  if (markdown.length > MAX_PARSABLE_LENGTH) {
+function parseExpensiMark(markdown: string, maxLength = MAX_PARSABLE_LENGTH): MarkdownRange[] {
+  if (markdown.length > maxLength) {
     return [];
   }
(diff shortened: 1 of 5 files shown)


Uzaifm127:

Uzaifm127:
@linhvovan29546 

Do we need the QA for this PR? 

I think we don't need it. We can test this in `react-native-live-markdown/example` app by modifying the `MAX_PARSABLE_LENGTH` from code. We can do the QA in version bump PR of `Expensify/App` repo.

---

@JS00001 

Do we need to update the `README.md` for this change? like the following:

```txt
The built-in `parseExpensiMark` parser accepts an optional `maxLength` argument. It defaults to 4,000 characters, so existing callers keep the current bounded behavior. Applications that have a different safe limit can pass that limit explicitly from a worklet parser:

function parser(input: string) {
  'worklet';

  return parseExpensiMark(input, 10000);
}
```

linhvovan29546:
> Do we need to update the README.md for this change? like the following:

I think we should

src/parseExpensiMark.ts
```diff
@@ -238,8 +238,8 @@ function parseTreeToTextAndRanges(tree: StackItem): [string, MarkdownRange[]] {
   return [text, ranges];
 }
 
-function parseExpensiMark(markdown: string): MarkdownRange[] {
-  if (markdown.length > MAX_PARSABLE_LENGTH) {
+function parseExpensiMark(markdown: string, maxLength = MAX_PARSABLE_LENGTH): MarkdownRange[] {
```
linhvovan29546, line 248: I think adding JSDoc would be helpful.

```
/**
 * Parses Markdown text into formatting ranges using ExpensiMark.
 *
 * @param markdown - The raw Markdown string to parse.
 * @param maxLength - Maximum input length to parse; longer input returns `[]` without parsing. Defaults to 4000.
 * @returns The list of formatting ranges, or `[]` if input exceeds `maxLength` or fails to parse.
 */
```

linhvovan29546 (approved):
A minor NIT https://github.com/Expensify/react-native-live-markdown/pull/775#issuecomment-5534432461. All good

src/parseExpensiMark.ts
Uzaifm127: Done.

Uzaifm127:
Updated the `README.md` file

linhvovan29546 (approved):

linhvovan29546:
@JS00001 All yours.

war-in (approved):

JS00001 (approved):

JS00001:
@Uzaifm127 I think we need one more PR to use the new version, right?

Uzaifm127:
@JS00001 Yes we need one more App PR in which we will be doing the version bump and passing the `MAX_MARKUP_LENGTH` as limit to `parseExpensiMark` function.

PR #776, opened 2026-09-08
[HOLD for RN 0.88] fix: keep the parser registration alive across effect remounts

dariusz-biela:
> [!IMPORTANT]
> **On hold until React Native 0.88.** The Fabric renderer in React Native 0.85–0.87 does not run the `useInsertionEffect` cleanup for a component removed while its `<Activity>` is hidden (see "Native renderer caveat"). React 19.3, which React Native 0.88 ships, always runs it. The diff itself does not need to change.

### Details

`MarkdownTextInput` registered its parser worklet during render and unregistered it in an effect cleanup. React can run that cleanup without unmounting the component (StrictMode, a hidden `<Activity>`), after which the input silently stopped formatting markdown.

**The fix**

The parser id is now chosen in JS, one per mounted input and a fresh one on every `parser` change, and the worklet is registered in `useInsertionEffect` inside the new `useParserId` hook. Insertion effects are neither double-invoked in StrictMode nor disconnected for a hidden `<Activity>`, and they run before the host tree is committed, so the registration lives exactly as long as the input and the first commit already carries a valid id.

The native registry becomes a plain map from id to worklet; an unknown id yields no ranges instead of throwing.

**Native renderer caveat**

React Native 0.85–0.87 build the Fabric renderer with `enableHiddenSubtreeInsertionEffectCleanup` off, so removing an input behind a hidden `<Activity>` skips the insertion cleanup and leaks its worklet in the native registry. Formatting stays correct because ids are never reused. `react-dom` has the flag on, and React 19.3 removed it and always runs the cleanup (facebook/react#30954, #34372, #35918).

**Example app**

New buttons reproduce the scenarios on a device: hide and show an `<Activity>` around the input, keep the input painted while hidden (`AlwaysPaintedView`), and swap the `parser` prop.

### Related Issues

https://github.com/Expensify/App/issues/98254

### Manual Tests

`src/__tests__/parserRegistration.test.tsx` covers mount, unmount, StrictMode, hidden and removed `<Activity>`, an abandoned render, shared parsers and a parser swap, and checks that nothing stays registered. The StrictMode and `<Activity>` cases fail without the fix.

Example app on iOS and Android with rebuilt native code:

1. Cold start: the initial value is formatted on the first frame.
2. Type `Hello *bold* and https://expensify.com`: bold and link render as you type.
3. Hide and show the `<Activity>`: the input comes back formatted and keeps formatting new text.
4. Swap the parser back and forth: the input follows each swap.
5. Wrap the input in `<React.StrictMode>` and type `*sm*`: it renders bold.

| iOS | Android |
| --- | --- |
| https://github.com/user-attachments/assets/399d4133-ca5a-47f6-bb91-56b9273ad0e1 | https://github.com/user-attachments/assets/41731a94-e59f-4ec7-9323-7a28e35727d4 |

### Linked PRs

https://github.com/Expensify/App/pull/100318

### DIFF
--- a/android/src/main/cpp/MarkdownParser.cpp
+++ b/android/src/main/cpp/MarkdownParser.cpp
@@ -7,14 +7,46 @@ using namespace facebook;
 
 namespace expensify {
 namespace livemarkdown {
+  jni::local_ref<MarkdownParser::jhybriddata> MarkdownParser::initHybrid(jni::alias_ref<jclass>) {
+    return makeCxxInstance();
+  }
+
+  void MarkdownParser::nativeSetParserId(const int parserId) {
+    std::unique_lock<std::mutex> lock(mutex_);
+    if (parserId_ == parserId) {
+      return;
+    }
+    const auto markdownWorklet = findMarkdownWorklet(parserId);
+    if (markdownWorklet == nullptr) {
+      return;
+    }
+    parserId_ = parserId;
+    markdownWorklet_ = markdownWorklet;
+  }
+
+  // A parse for the current id uses the worklet kept alive by `nativeSetParserId`.
+  // Any other id is looked up in the registry the way it always was.
+  std::shared_ptr<SerializableWorklet> MarkdownParser::workletForParserId(const int parserId) {
+    {
+      std::unique_lock<std::mutex> lock(mutex_);
+      if (parserId_ == parserId) {
+        return markdownWorklet_;
+      }
+    }
+
+    return findMarkdownWorklet(parserId);
+  }
+
   jni::local_ref<jni::JString> MarkdownParser::nativeParse(
-      jni::alias_ref<jhybridobject> jThis,
       jni::alias_ref<jni::JString> text,
       const int parserId) {
-    const auto markdownRuntime = expensify::livemarkdown::getMarkdownRuntime();
-    jsi::Runtime &rt = markdownRuntime->getJSIRuntime();
+    const auto markdownWorklet = workletForParserId(parserId);
+    if (markdownWorklet == nullptr) {
+      return jni::make_jstring("[]");
+    }
 
-    const auto markdownWorklet = expensify::livemarkdown::getMarkdownWorklet(parserId);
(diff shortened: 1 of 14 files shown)


dariusz-biela:

android/src/main/cpp/MarkdownParser.cpp
```diff
@@ -11,11 +11,14 @@ namespace livemarkdown {
       jni::alias_ref<jhybridobject> jThis,
       jni::alias_ref<jni::JString> text,
       const int parserId) {
+    const auto markdownWorklet = expensify::livemarkdown::findMarkdownWorklet(parserId);
+    if (markdownWorklet == nullptr) {
+      return jni::make_jstring("[]");
```
mateuuszzzzz: Could we distinguish "no worklet registered" from "parser returned zero ranges", skip the cache write in the first case, and log a warning with the `parserId`? That would make the broken invariant visible. What do you think?

mateuuszzzzz: By "broken invariant" I mean the parser is registered in `useInsertionEffect`, so it should always be available when the native view calls parse

dariusz-biela: Done in 327efd2: native returns null / nil for an unknown id, logs a warning with the parserId and leaves the cache alone (both platforms). On iOS the warmup also skips its completion in that case, otherwise the main thread would loop on re-measure.

One known false positive: the worklet is unregistered on the JS thread during the commit, but the native view gets the new `parserId` (or is removed) only when the UI thread runs the mount transaction. A parse in between (a keystroke, or an iOS warmup scheduled just before) still uses the old id and logs the warning. It is a few-ms window, only after a parser swap or an unmount, and harmless: nothing is cached and the view reformats once the new id lands. I did not suppress it on purpose - remembering unregistered ids would also hide a cleanup that runs without an unmount, which is the bug this PR fixes. If the noise matters I can log once per parserId instead.

